PR #275, opened 2020-09-14
Disable Strict SSL for On-Prem Providers

cstryker:
Use custom agent for on-prem providers when strict SSL is disabled

**This PR Addresses:**  
[Agent is rejecting responses from on-prem services (bitbucket, jira, ...) when they use self-signed certs or certs signed by their own CA](https://trello.com/c/hPZCpzuV/4702-agent-is-rejecting-responses-from-on-prem-services-bitbucket-jira-when-they-use-self-signed-certs-or-certs-signed-by-their-own-c)


cstryker:
@mfarias&nbsp;nudge

PR #319, opened 2020-11-10
prune, compactify, and debounce updating modifiedRepos

cstryker:
Note that until we can ensure that all users have this update, as well as the API server updated needed to support it, we have to store _both_ the modifiedRepos structure, and the compactified version. I have worked to "prune" the modifiedRepos structure, which results in some improvement to the size, but compactified is much better.

**This PR Addresses:**  
[Investigate large modifiedRepos objects and what to do about them](https://trello.com/c/So7GKCT8/4881-investigate-large-modifiedrepos-objects-and-what-to-do-about-them)

### DIFF
--- a/shared/agent/src/api/apiProvider.ts
+++ b/shared/agent/src/api/apiProvider.ts
@@ -130,7 +130,6 @@ import {
 	SetCodemarkStatusRequest,
 	SetCodemarkStatusResponse,
 	SetModifiedReposRequest,
-	SetModifiedReposResponse,
 	SetStreamPurposeRequest,
 	SetStreamPurposeResponse,
 	ThirdPartyProviderSetTokenRequest,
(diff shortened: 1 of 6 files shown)


marcelofarias (approved):

bcanzanella:
this removed code is ok? (just checking)

```
const _updateModifiedRepos = (modifiedRepos: RepoScmStatus[], teamId: string) => async (
	dispatch,
	getState: () => CodeStreamState
) => {
	return HostApi.instance.send(SetModifiedReposRequestType, {
		modifiedRepos,
		teamId
	});
};
```
								
[shared/ui/store/users/actions.ts](https://github.com/teamcodestream/codestream/blob/21fd437be1659ea0f385d8b37e364b66e0dacae0/shared/ui/store/users/actions.ts#L60-L68) (Lines 60-68)

cstryker:
@brian&nbsp;yes, though what you highlighted is the whole function, but only a portion of that function has been removed, in case that's not clear ... the reason is that now that saving modifiedRepos is debounced, there isn't a guarantee of immediate feedback when modifiedRepos is changed, so we can't rely on this method to update the UI right away ... the debounce is such that the Pubnub message that comes back in response to the update should update the UI, but this may take as long as a minute if there was an update just prior ... i have tested and it does work as expected

bcanzanella (approved):

PR #326, opened 2020-11-16
Fix "failure to recover from agent crash"

cstryker:
After an agent crash, properly restart and wait for initialization before resetting user state

**This PR Addresses:**  
[Failure to recover from agent crash](https://trello.com/c/FiUOhvro/5188-failure-to-recover-from-agent-crash)


cstryker:
Note: a similar fix may need to be implemented for VS, JB, and (ugh) Atom

bcanzanella (approved):
wow, nice, this is some serious getting-into-the-weeds stuff.

PR #340, opened 2020-11-25
support a banner if bad github token is detected, allow to reauth or disconnect

cstryker:
**This PR Addresses:**  
[Implement banner for bad PR-related access tokens](https://trello.com/c/ZksWKRSX/5234-implement-banner-for-bad-pr-related-access-tokens)

### DIFF
--- a/shared/agent/src/providers/github.ts
+++ b/shared/agent/src/providers/github.ts
@@ -180,7 +180,28 @@ export class GitHubProvider extends ThirdPartyIssueProviderBase<CSGitHubProvider
 	};
 
 	async query<T = any>(query: string, variables: any = undefined) {
-		const response = await (await this.client()).request<any>(query, variables);
+		if (this._providerInfo && this._providerInfo.tokenError) {
+			throw new Error("Access token invalid");
+		}
+
+		let response;
+		try {
+			response = await (await this.client()).request<any>(query, variables);
+		}
+		catch (ex) {
+			if (ex.response && ex.response.message === "Bad credentials") {
+				this.session.api.setThirdPartyProviderInfo({
+					providerId: this.providerConfig.id,
+					data: {
+						tokenError: {
+							error: ex,
+							occurredAt: Date.now()
+						}
+					}
+				});
+			}
+			throw ex;
+		}
 
 		try {
 			if (Logger.level === TraceLevel.Debug && response && response.rateLimit) {
--- a/shared/agent/src/providers/github.ts
+++ b/shared/agent/src/providers/github.ts
@@ -6,6 +6,7 @@ import * as paths from "path";
 import * as qs from "querystring";
 import { CodeStreamSession } from "session";
 import { URI } from "vscode-uri";
+import { InternalError, ReportSuppressedMessages } from "../agentError";
 import { SessionContainer } from "../container";
 import { Logger, TraceLevel } from "../logger";
 import {
@@ -181,7 +182,7 @@ export class GitHubProvider extends ThirdPartyIssueProviderBase<CSGitHubProvider
 
 	async query<T = any>(query: string, variables: any = undefined) {
 		if (this._providerInfo && this._providerInfo.tokenError) {
-			throw new Error("Access token invalid");
+			throw new InternalError(ReportSuppressedMessages.AccessTokenInvalid);
 		}
 
 		let response;
@@ -200,7 +201,7 @@ export class GitHubProvider extends ThirdPartyIssueProviderBase<CSGitHubProvider
 					}
 				});
 			}
-			throw ex;
+			throw new InternalError(ReportSuppressedMessages.AccessTokenInvalid, { error: ex });
 		}
 
 		try {
--- a/shared/agent/src/providers/provider.ts
+++ b/shared/agent/src/providers/provider.ts
@@ -484,6 +484,10 @@ export abstract class ThirdPartyProviderBase<
 		init: RequestInit,
 		options: { [key: string]: any } = {}
 	): Promise<ApiResponse<R>> {
+		if (this._providerInfo && this._providerInfo.tokenError) {
+			throw new Error("Access token invalid");
+		}
+
 		const start = process.hrtime();
 
 		let traceResult;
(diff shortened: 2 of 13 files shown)


shared/ui/Stream/PRProviderErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+			supportsReauth: capabilities.providerReauth,
+			offline: connectivity.offline,
+			failedProvider: failedProviderId ? providers[failedProviderId] : null,
+			tokenError: tokenError.accessTokenError
+		};
+	});
+	
+	const onClickReauthorize = event => {
+		event.preventDefault();
+		dispatch(connectProvider(derivedState.failedProvider!.id, "Provider Error Banner", true));
```
bcanzanella, line 56: this might have to be `configureAndConnectProvider` rather than `connectProvider` to deal with enterprise configs where we need a PAT, otherwise nothing will happen for connections like GHE.

shared/agent/src/providers/github.ts
```diff
@@ -180,7 +180,28 @@ export class GitHubProvider extends ThirdPartyIssueProviderBase<CSGitHubProvider
 	};
 
 	async query<T = any>(query: string, variables: any = undefined) {
-		const response = await (await this.client()).request<any>(query, variables);
+		if (this._providerInfo && this._providerInfo.tokenError) {
+			throw new Error("Access token invalid");
```
bcanzanella, line 184: I'd want to make sure that this isn't captured by Sentry.

shared/agent/src/providers/github.ts
```diff
@@ -180,7 +180,28 @@ export class GitHubProvider extends ThirdPartyIssueProviderBase<CSGitHubProvider
...
+					providerId: this.providerConfig.id,
+					data: {
+						tokenError: {
+							error: ex,
+							occurredAt: Date.now()
+						}
+					}
+				});
+			}
+			throw ex;
```
bcanzanella, line 203: Same here.

shared/agent/src/providers/provider.ts
```diff
@@ -484,6 +484,10 @@ export abstract class ThirdPartyProviderBase<
 		init: RequestInit,
 		options: { [key: string]: any } = {}
 	): Promise<ApiResponse<R>> {
+		if (this._providerInfo && this._providerInfo.tokenError) {
```
bcanzanella, line 487: Here too?

shared/ui/store/providers/reducer.ts
```diff
@@ -122,15 +125,21 @@ export const isConnectedSelectorFriendly = (
 					info != undefined &&
 					info.hosts != undefined &&
 					Object.keys(info.hosts).some(host => {
-						return providers[host] != undefined && info.hosts![host].accessToken != undefined;
+						const isConnected = providers[host] != undefined && info.hosts![host].accessToken != undefined;
+						if (isConnected && accessTokenError) {
+							accessTokenError.accessTokenError = info.hosts![host].tokenError;
```
bcanzanella, line 130: i'm not totally sure what the assignments to `accessTokenError` are for here and the few below?

bcanzanella:
I'd also consider adding something in registry.ts (could work here: https://api.codestream.com/c/W75JY6zGmBhgt48D/POHQrTmzS-m7s_ymyA9T5w) to exclude any user providers that have the `accessTokenError` so there's no unnecessary polling.

shared/ui/Stream/PRProviderErrorBanner.tsx
cstryker: good point

shared/agent/src/providers/github.ts
cstryker: yeah, that _was_ the whole point, wasn't it? :)

shared/ui/store/providers/reducer.ts
cstryker: i'll add some commentary to make it clear

shared/ui/Stream/PRProviderErrorBanner.tsx
cstryker: fixed

shared/agent/src/providers/github.ts
cstryker: fixed

shared/agent/src/providers/github.ts
cstryker: fixed

shared/agent/src/providers/provider.ts
cstryker: fixed

shared/ui/store/providers/reducer.ts
cstryker: see the comments, if still not clear, i'll explain

cstryker:
> I'd also consider adding something in registry.ts (could work here: https://api.codestream.com/c/W75JY6zGmBhgt48D/POHQrTmzS-m7s_ymyA9T5w) to exclude any user providers that have the `accessTokenError` so there's no unnecessary polling.

not sure i agree with this ... i don't think we want to make the assumption that `getConnectedProviders` should return only connected providers for which the access token is known to be valid ... i think that now that the sentry errors will be suppressed, it is ok to keep polling, since the polling just results in an exception thrown ... to stop the polling, we'd need a mechanism to resume it once the access token is fixed

shared/ui/store/providers/reducer.ts
bcanzanella: gotcha, it's acting like `ref` or `out` in c#.

bcanzanella:
> 
> 
> > I'd also consider adding something in registry.ts (could work here: https://api.codestream.com/c/W75JY6zGmBhgt48D/POHQrTmzS-m7s_ymyA9T5w) to exclude any user providers that have the `accessTokenError` so there's no unnecessary polling.
> 
> not sure i agree with this ... i don't think we want to make the assumption that `getConnectedProviders` should return only connected providers for which the access token is known to be valid ... i think that now that the sentry errors will be suppressed, it is ok to keep polling, since the polling just results in an exception thrown ... to stop the polling, we'd need a mechanism to resume it once the access token is fixed

ah, i should have said `no unnecessary polling _action_`. meaning, the polling would always run, it's just that it wouldn't find any valid providers for that interval and just return. in any case, what you have is fine.

bcanzanella (approved):

shared/ui/store/providers/reducer.ts
cstryker: yup

PR #344, opened 2020-11-30
Feature/multiline code fences

bcanzanella:
`excludeParagraphWrap` does exclude the `<p>` tag, but it also prevents html highlighting (prettyPrint) from working

### DIFF
--- a/shared/ui/Stream/Codemark.tsx
+++ b/shared/ui/Stream/Codemark.tsx
@@ -332,15 +332,14 @@ export class Codemark extends React.Component<Props, State> {
 
 	renderTextReplaceCodeBlocks = (text: string) => {
 		const { codemark, capabilities } = this.props;
-		if (!codemark || !codemark.markers)
-			return <MarkdownText text={text} excludeParagraphWrap={true} />;
+		if (!codemark || !codemark.markers) return <MarkdownText text={text} inline={true} />;
 
 		const blocks: any[] = [];
 		const groups = text.split(/\[#(\d+)]/);
 		let index = 0;
 		this.skipMarkers = [];
 		while (index < groups.length) {
-			blocks.push(<MarkdownText text={groups[index]} excludeParagraphWrap={false} />);
+			blocks.push(<MarkdownText text={groups[index]} />);
 			if (index + 1 < groups.length) {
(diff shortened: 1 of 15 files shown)


shared/ui/utils.ts
```diff
@@ -384,6 +384,9 @@ export function escapeHtml(text: string) {
  */
 export function replaceHtml(text: string) {
 	const domParser = new DOMParser();
+	//input text's newlines will be created with <div> or <br> tags
+	//remove extra \n or \r\n to remove double lines later in markdown
+	text = text.replace(/\r\n/g, "").replace(/\n/g, "");
```
bcanzanella, line 389: this fixes the weird "dance" that previewing markdown would do, where additional empty lines would be inserted.

shared/ui/Stream/Codemark.tsx
```diff
@@ -340,7 +340,7 @@ export class Codemark extends React.Component<Props, State> {
 		let index = 0;
 		this.skipMarkers = [];
 		while (index < groups.length) {
-			blocks.push(<MarkdownText text={groups[index]} excludeParagraphWrap={true} />);
+			blocks.push(<MarkdownText text={groups[index]} excludeParagraphWrap={false} />);
```
bcanzanella, line 343: while `excludeParagraphWrap={true}` does remove the wrapping `<p>` tag, it also means use the `renderInline` function which prevents the html highlighting from working. 

To make this clearer, i was thinking of renaming that property to be named `inline`.

thoughts?

ppezaris: sounds reasonable to me... might be worth adding a comment as documentation

ppezaris: for this change, why not just remove excludeParagraphWrap=[whatever] since it defaults to false?

ppezaris:
other than my comments LGTM

shared/ui/Stream/Codemark.tsx
bcanzanella: sounds good, I'll do that property rename and remove the `excludeParagraphWrap={false}` instances

ppezaris (approved):

PR #359, opened 2020-12-17
git remote ssh alias

bcanzanella:
you can recreate this with the following setup...

there are some github & gitlab cloud users that have ssh aliases setup, including some CodeStream users. this PR addresses those users. With a setup like below, you can replicate the issues these users have faced. 

1) This prevents us from certain scenarios where we try to figure out which provider is tied to a remote -- there's some logic that checks if `domain === github.com` and `domain === gitlab.com` -- which will fail in these aliased cases.
2) We end up storing bad data to Mongo. In the case of my beer & cheese remote, we're currently storing a normalizedUrl that looks like `beer.github.com/TeamCodeStream/codestream` which is invalid. it should be `github.com/TeamCodeStream/codestream`

remote...
```
origin  git@beer.github.com-cheese:TeamCodeStream/codestream.git (fetch)
origin  git@beer.github.com-cheese:TeamCodeStream/codestream.git (push)
```

and ssh config entry...

```
Host beer.github.com-cheese
	HostName github.com
	User brian
	IdentityFile ~/.ssh/id_rsa
```

local .git/config

```
[remote "origin"]
	url = git@beer.github.com-cheese:TeamCodeStream/codestream.git
	fetch = +refs/heads/*:refs/remotes/origin/*
```

My main question is to see if there's a scenario where `ssh -G` would fail or be unavailable? it's guarded and will fallback to the current logic if it fails.

### DIFF
--- a/shared/agent/src/git/parsers/remoteParser.ts
+++ b/shared/agent/src/git/parsers/remoteParser.ts
@@ -1,13 +1,15 @@
 "use strict";
 import { GitRemote, GitRemoteType } from "../models/remote";
+import * as childProcess from "child_process";
+import { Logger } from "../../logger";
 
 const emptyStr = "";
 
 const remoteRegex = /^(.*)\t(.*)\s\((.*)\)$/gm;
 const urlRegex = /^(?:(git:\/\/)(.*?)(?::.*?)?\/|(https?:\/\/)(?:.*?@)?(.*?)(?::.*?)?\/|git@(.*):|(ssh:\/\/)(?:.*@)?(.*?)(?::.*?)?(?:\/|(?=~))|(?:.*?@)(.*?):)(.*)$/;
-
+const hostnameRegex = new RegExp("hostname (.*)");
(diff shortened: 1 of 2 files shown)


shared/agent/src/git/parsers/remoteParser.ts
```diff
@@ -56,20 +58,51 @@ export class GitRemoteParser {
 		return remotes;
 	}
 
-	static parseGitUrl(url: string): [string, string, string] {
-		const match = urlRegex.exec(url);
-		if (match == null) return [emptyStr, emptyStr, emptyStr];
-
-		let host = match[5];
-		const githubHost = "github.com";
```
bcanzanella, line 64: this is somewhat faulty, and it will work for someone that aliases github.com with a suffix, but not when it has a prefix, or even when it's "pineapple.com" (which could be an alias for github.com)

marcelofarias (approved):

PR #451, opened 2021-03-19
Feature: unreads

ppezaris:


shared/ui/store/users/reducer.ts
```diff
@@ -154,6 +156,60 @@ export const getPreferences = createSelector(
...
+	(a: any, item: Readable) => item,
+	(umis: UnreadsState | undefined, item: Readable) => {
+		if (!umis || !item) return false;
+		const { lastReadItems } = umis;
+		return isItemUnread(item, lastReadItems[item.id]);
+	}
+);
+
+export const unreadMap = createSelector(
+	(state: CodeStreamState) => (isFeatureEnabled(state, "readItem2") ? state.umis : undefined),
```
cstryker, line 194: what's this?

shared/ui/Stream/Codemark.tsx
```diff
@@ -2109,6 +2124,8 @@ const mapStateToProps = (state: CodeStreamState, props: InheritedProps): Connect
 			? getReview(state.reviews, codemark.reviewId)
 			: undefined;
 
+	const unread = isUnread(state, codemark!);
+	console.warn("UNREAD IS: ", unread, " for codemark ", codemark);
```
cstryker, line 2128: remove warn?

shared/ui/Stream/Codemark.tsx
```diff
@@ -160,6 +160,9 @@ interface InheritedProps {
 
 type Props = InheritedProps & DispatchProps & ConnectedProps;
 
+const EMPTY_OBJECT = {};
+const EMPTY_OBJECT2 = {};
```
cstryker, line 164: what's this for?

shared/ui/utils.ts
```diff
@@ -7,6 +7,7 @@ export interface Disposable {
 	dispose(): void;
 }
 
+// DO NOT USE THESE
```
cstryker, line 10: not that i would, but why?

cstryker:
other than minor comments, LGTM

shared/ui/utils.ts
ppezaris: because there global, and if they're used in multiple places you might have unintended consequences if they get set.

PR #452, opened 2021-03-23
roll off some of the GH checks

bcanzanella:
changes some of the polls from every 1 min to every 2 min

### DIFF
--- a/shared/ui/Stream/PullRequest.tsx
+++ b/shared/ui/Stream/PullRequest.tsx
@@ -531,11 +531,11 @@ export const PullRequest = () => {
 				_checkMergeabilityStatus().then(_ => {
 					setAutoCheckedMergeability(_ ? "CHECKED" : "UNKNOWN");
 				});
-			}, 5000);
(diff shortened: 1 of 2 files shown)


shared/ui/Stream/PullRequest.tsx
```diff
@@ -531,11 +531,11 @@ export const PullRequest = () => {
...
 				});
-			}, 5000);
+			}, 8000);
 		}
 		interval = setInterval(async () => {
-			// checks for 1 hour
-			if (intervalCounter >= 60) {
+			// checks for 15 min
+			if (intervalCounter >= 3) {
 				interval && clearInterval(interval);
```
ppezaris, line 539: checks for 15 minutes and then stops checking?

bcanzanella: yup, but that one is only when you are looking at the PR. 
maybe that's not the correct interval + timeout. maybe a wider interval but longer timespan would be better?

PR #617, opened 2021-06-10
Removed live-view paywall

patrick-morgan:
Changed it so everyone can see the live-view (no longer paywalled). Free teams can rejoice, hoorah!


bcanzanella (approved):
looks good!

PR #633, opened 2021-06-22
Fix start review

patrick-morgan:
There was a bug where if you tried to  start a review on a PR by commenting on the  code and you weren't selecting "all commits" you weren't able to start a review.

The issue was that when you weren't selecting all commits the textEditorUriContext.headbranch and basebranch were both the branch name (for example 'my-cool-branch'). But, if you were selecting all commits the basebranch would be 'main' and the headBranch would be the base branch (e.g. 'my-cool-branch') therefore it was entering the wrong case in the switch statement.

Therefore, switching the order of the cases sets the prRange to the, what I believe to be, correct values.


bcanzanella (approved):
👍🏻

PR #635, opened 2021-06-22
Feature/updated pr doesnt2

patrick-morgan:
Added functionality so that when a user changes an item on the pullrequestconversationtab pane we refresh the user's prs asynchronously so that when they exit out of the pr their pull requests are up to date for each query listed.

Currently using a timeout to delay the refreshing of the prs long enough for GH's servers to process the update (e.g. adding a label, assignee, etc), the timeout is set to 2 seconds but this means that sometimes the data will be refreshed too early and not be displayed properly. There is surely a better way to do this.

### DIFF
--- a/shared/ui/Stream/OpenPullRequests.tsx
+++ b/shared/ui/Stream/OpenPullRequests.tsx
@@ -404,7 +404,6 @@ export const OpenPullRequests = React.memo((props: Props) => {
 		// need to check if it was new/editing pullRequestQueries or just updating other preferences
 		if (!isEqual(queries, newQueries)) {
 			setQueries(newQueries);
-			// update store
 		}
 	}, [derivedState.pullRequestQueries]);
 
--- a/shared/ui/Stream/OpenPullRequests.tsx
+++ b/shared/ui/Stream/OpenPullRequests.tsx
@@ -25,7 +25,7 @@ import {
 } from "@codestream/protocols/agent";
 import { OpenUrlRequestType, WebviewPanels } from "@codestream/protocols/webview";
 import { Button } from "../src/components/Button";
-import { getMyPullRequests, openPullRequestByUrl, updatePullRequestGroups } from "../store/providerPullRequests/actions";
+import { getMyPullRequests, openPullRequestByUrl } from "../store/providerPullRequests/actions";
 import { PRBranch } from "./PullRequestComponents";
 import { PRHeadshotName } from "../src/components/HeadshotName";
 import styled from "styled-components";
@@ -211,7 +211,6 @@ export const OpenPullRequests = React.memo((props: Props) => {
 		const prConnectedProvidersWithErrors = prConnectedProviders.filter(_ => _.hasAccessTokenError);
 		const prConnectedProvidersLength = prConnectedProviders.length;
 		const myPullRequests = getMyPullRequestsSelector(state);
-		const pullRequestGroups = state.providerPullRequests.pullRequestGroups;
 
 		return {
 			repos,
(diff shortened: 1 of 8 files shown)


shared/ui/Stream/OpenPullRequests.tsx
```diff
@@ -394,6 +404,7 @@ export const OpenPullRequests = React.memo((props: Props) => {
 		// need to check if it was new/editing pullRequestQueries or just updating other preferences
 		if (!isEqual(queries, newQueries)) {
 			setQueries(newQueries);
+			// update store
```
bcanzanella, line 407: is this comment old, or is this a real TODO?

shared/ui/store/providerPullRequests/reducer.ts
```diff
@@ -689,9 +707,20 @@ export function reduceProviderPullRequests(
 			}
 			return {
 				myPullRequests: { ...state.myPullRequests },
-				pullRequests: newState
+				pullRequests: newState,
 			};
 		}
+		// case ProviderPullRequestActionsTypes.ClearPullRequestError: {
```
bcanzanella, line 713: was this removed for testing?

shared/ui/store/providerPullRequests/reducer.ts
```diff
@@ -50,7 +50,7 @@ export function reduceProviderPullRequests(
 				newState[action.payload.providerId][id].error = undefined;
 				return {
 					myPullRequests: { ...state.myPullRequests },
-					pullRequests: newState
+					pullRequests: newState,
```
bcanzanella, line 53: nit pick on these trailing commas (there are a bunch in this file). if you have the esbenp.prettier-vscode extension setup it should remove them all for you. (I like to enable format on save)

shared/ui/Stream/PullRequestConversationTab.tsx
```diff
@@ -590,13 +624,45 @@ export const PullRequestConversationTab = (props: {
...
+		});
+		fetchPRs();
+	};
+
+	const fetchPRs = async () => {
+		// We need to delay these api requests as we are making a request to make a change (e.g. adding a label)
+		// then refreshing the new data and it takes some time for the back-end to do this
+		for (const connectedProvider of derivedState.PRConnectedProviders) {
+			try {
+				if (queries) {
```
bcanzanella, line 647: can this be `derivedState.pullRequestQueries`?

shared/ui/Stream/PullRequestConversationTab.tsx
```diff
@@ -590,13 +624,45 @@ export const PullRequestConversationTab = (props: {
...
+					const options = { force: true, alreadyLoading: false };
+					const providerQuery: PullRequestQuery[] = queries[connectedProvider.id];
```
bcanzanella, line 649: there are a few other spots where we're defaulting the queries with `|| defaultQueries[connectedProvider.id]]` and i think we should do that here too. (queries don't exist in a user object if they haven't modified them)

bcanzanella (changes requested):
Couple comments

shared/ui/Stream/OpenPullRequests.tsx
patrick-morgan: old comment

shared/ui/store/providerPullRequests/reducer.ts
patrick-morgan: had  it moved down there for testing, it is still above not commented out. Deleted the commented out code

shared/ui/store/providerPullRequests/reducer.ts
patrick-morgan: linted file 👍

shared/ui/Stream/PullRequestConversationTab.tsx
patrick-morgan: Good catch, changed this to use defaultQueries. If we want to store defaultQueries in the store then I can add that too

shared/ui/store/providerPullRequests/reducer.ts
bcanzanella: ah gotcha, thanks. that's what i get for starting this review on gh.com (no context of the whole file) :/

shared/ui/Stream/PullRequestConversationTab.tsx
bcanzanella: i think what you have will work!

bcanzanella (approved):

shared/ui/Stream/PullRequestConversationTab.tsx
patrick-morgan: yep

PR #669, opened 2021-07-26
Feature/update gitlab search mrs

patrick-morgan:
-changed mr queries to match gitlab syntax
-updated filtering modals/examples
-fixed bug where prs were cleared out when testing a custom filter
-updated default queries
-added backwards compatibility support for old filter syntax


bcanzanella (approved):
:+1:

PR #717, opened 2021-09-15
Gitlab mr issues

patrick-morgan:
Fixed default setting for deleting source branch, and added squash_option support for Gitlab MRs.

### DIFF
--- a/shared/agent/src/protocol/agent.protocol.gitlab.ts
+++ b/shared/agent/src/protocol/agent.protocol.gitlab.ts
@@ -61,6 +61,7 @@ export interface GitLabMergeRequestWrapper {
 		onlyAllowMergeIfPipelineSucceeds: boolean;
 		allowMergeOnSkippedPipeline: boolean;
 		onlyAllowMergeIfAllDiscussionsAreResolved: boolean;
+		squashOption: string;
 		/*
 		merge:
 		Merge commit
--- a/shared/agent/src/protocol/agent.protocol.gitlab.ts
+++ b/shared/agent/src/protocol/agent.protocol.gitlab.ts
@@ -61,7 +61,7 @@ export interface GitLabMergeRequestWrapper {
 		onlyAllowMergeIfPipelineSucceeds: boolean;
 		allowMergeOnSkippedPipeline: boolean;
 		onlyAllowMergeIfAllDiscussionsAreResolved: boolean;
-		squashOption: string;
+		squashOption: "never" | "always" | "default_on" | "default_off";
 		/*
 		merge:
 		Merge commit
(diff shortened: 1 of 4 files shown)


shared/ui/Stream/PullRequests/GitLab/MergeBox.tsx
```diff
@@ -374,39 +381,57 @@ export const MergeBox = props => {
...
-							</Link>
-						</div>
+						{derivedState.prRoot && derivedState.prRoot.project.squashOption === "always" ? (
+							<div className="pad-left">
+								<Checkbox
+									checked={true}
+									name="squash"
+									noMargin
+									onChange={() => {}}
+									disabled=" "
```
bcanzanella, line 403: remove the "disabled" field

shared/agent/src/protocol/agent.protocol.gitlab.ts
```diff
@@ -61,6 +61,7 @@ export interface GitLabMergeRequestWrapper {
 		onlyAllowMergeIfPipelineSucceeds: boolean;
 		allowMergeOnSkippedPipeline: boolean;
 		onlyAllowMergeIfAllDiscussionsAreResolved: boolean;
+		squashOption: string;
```
bcanzanella, line 64: if there is a known enum of values here, you can use those as the type:

`squashOption: "default_on" | "always" | "never"`

shared/agent/src/providers/gitlab.ts
```diff
@@ -1333,6 +1331,7 @@ export class GitLabProvider extends ThirdPartyIssueProviderBase<CSGitLabProvider
 				only_allow_merge_if_all_discussions_are_resolved: boolean;
 				only_allow_merge_if_pipeline_succeeds: boolean;
 				allow_merge_on_skipped_pipeline: boolean;
+				squash_option: string;
```
bcanzanella, line 1334: same w/r/t the squash type not being a string. you could create a TypeScript type that has all the enums (so that it's reusable) and can be reused here.

shared/ui/Stream/PullRequests/GitLab/MergeBox.tsx
```diff
@@ -374,39 +381,57 @@ export const MergeBox = props => {
...
 							</Checkbox>
 						</div>
-						<div className="pl5">
-							<Link
-								href={`${derivedState.pr.baseWebUrl}/help/user/project/merge_requests/squash_and_merge`}
-							>
-								<Icon name="question" title="What is squashing?" placement="top" />
-							</Link>
-						</div>
+						{derivedState.prRoot && derivedState.prRoot.project.squashOption === "always" ? (
```
bcanzanella, line 396: i think you could wrap this whole section with a `derivedState.prRoot` check if you'd like to minimize multiple checks

```
{derivedState.prRoot && (
   <>
    {derivedState.prRoot.project.squashOption === "always" ? () : ()}
    {derivedState.prRoot.project.squashOption !== "never" && ()}
   </>
)}
```

bcanzanella (changes requested):

shared/ui/Stream/PullRequests/GitLab/MergeBox.tsx
patrick-morgan: without that the checkbox will appear to be clickable. I added the disabledEmpty field so that now the checkbox will be disabled but there won't display any text after that would mess up the formatting

bcanzanella (approved):

PR #786, opened 2021-12-06
CDSTRM-91: Preserve organization on sign-in

purplexa:
This adds storing the last selected organization in the user.preferences
object in the database as `lastTeamId` and checking that value on login
so that we maintain the last selected organization across logout/login
instead of just falling back to the oldest organization.


bcanzanella (approved):
looks great!

PR #794, opened 2021-12-21
CDSTRM-1301: Allow saving the "recent" PR query …

purplexa:
This enables users to save the default "recent" PR query, which is a special string we handle specifically. This allows them to recreate it if they've changed it.


bcanzanella (approved):
nice!

PR #803, opened 2022-01-05
CDSTRM-1332: Disable FROP by default for new users

purplexa:
This changes the default for new users for the feature which shows a toast notification prompting you to review unreviewed commits when you pull new code. New users are arbitrarily defined as anyone who joined after 2022-01-05T17:50:00Z, mostly because the number looked kind of nice and was appropriate to make it easy to test that it works. It does mean there will probably be a few users who join between that time and when this releases who get their value changed, but that's probably okay.


bcanzanella (approved):
nice!

PR #806, opened 2022-01-08
feature/CSTRM_1315_2

eric-j:
Spec: Simplified Onboarding Flows from NR1 (PHASE 1)

I wrote this rather quickly since we want to release next week and I am out.
Authentication and the flow that runs through CompanyCreation should work pretty well, but I would recommend QAing before a production release.


bcanzanella:
looks good! we're going to wait a bit to merge this.

bcanzanella (approved):

PR #903, opened 2022-03-10
Fix: CDSTRM 1116-theme-not-applied

dsellarsnr:


marcelofarias:
It looks good regarding theme changes. Any insights about initialization timing? This is harder to reproduce - not sure if there's any other events we could listen to to know the theme is ready.

PR #914, opened 2022-03-17
Feature/vseu

bcanzanella:


vs/src/CodeStream.VisualStudio/Services/AuthenticationService.cs
```diff
@@ -40,7 +40,7 @@ public class AuthenticationService : IAuthenticationService {
 		[Import]
 		public IWebviewUserSettingsService WebviewUserSettingsService { get; set; }
 
-		public async System.Threading.Tasks.Task LogoutAsync(SessionSignedOutReason reason = SessionSignedOutReason.UserSignedOutFromWebview, JToken payload = null) {
+		public async System.Threading.Tasks.Task LogoutAsync(SessionSignedOutReason reason = SessionSignedOutReason.UserSignedOutFromWebview, string newServerUrl = null, string newEnvironment = null, JToken payload = null) {
```
marcelofarias, line 43: I believe newEnvironment is not being used here.

marcelofarias (commented):
Just a minor comment about a newEnvironment parameter not being used. In case it does need to be used, let me know because I would need to add that to JB as well.

vs/src/CodeStream.VisualStudio/Services/AuthenticationService.cs
bcanzanella: should be alright asis

PR #936, opened 2022-03-29
Feature/react tests2

bcanzanella:
- upgrades react to 16.9.X
- adds back some react component tests
- adds shared/ui coverage reports (also in TC)
- adds coverage for agent (non-TC, mem issue)


shared/agent/package.json
```diff
@@ -33,13 +33,14 @@
 		"rebuild": "npm run clean && npm install --no-save && npm run build",
 		"update:emoji": "pushd emoji && node ./shortcodeToEmoji.js && popd",
 		"watch": "webpack --watch --env development=true",
+		"test:ci": "npm run test-unit",
```
dsellarsnr, line 36: Should this run the test-unit-with-coverage variant?

bcanzanella: yeah, good catch. there was an out of mem error on the CI server with it, but i've added another setting to fix it (last CI build looks ok)

PR #953, opened 2022-04-12
NR-818 Rename Clubhouse to Shortcut

dsellarsnr:
**This PR Addresses:**  
[NR-818 Rename Clubhouse to Shortcut](https://issues.newrelic.com/browse/NR-818)  


shared/ui/Stream/icons8-data.json
```diff
@@ -126,8 +126,8 @@
 		"name": "clubhouse",
```
bcanzanella, line 126: rename?

bcanzanella, line 126: rename? (not sure what this isn't on the right side of the diff...)

dsellarsnr: In mongo we have the providerInfo.clubhouse.accessToken stored as well as any custom filters - I thought it would be easier to keep the providerId as clubhouse so that we don't have to do a database migration. So, I kept the providerId and the related lookups like this one as clubhouse. But if a DB migration is easy we can do a full rename of everything.

bcanzanella: hmm, i'll let @cstryker chime in here.

jb/docs/description.html
```diff
@@ -7,7 +7,7 @@
 <h1 id="integrations">Integrations</h1>
 <ul>
     <li><b>Code Hosts:</b> Bitbucket, Bitbucket Server, GitHub, GitHub Enterprise, GitLab, GitLab Self-Managed</li>
-    <li><b>Issue Trackers:</b> Asana, Azure DevOps, Bitbucket, Clubhouse, GitHub, GitHub Enterprise, GitLab, GitLab Self-Managed, Jira, Linear, Trello, YouTrack</li>
+    <li><b>Issue Trackers:</b> Asana, Azure DevOps, Bitbucket, Shortcut, GitHub, GitHub Enterprise, GitLab, GitLab Self-Managed, Jira, Linear, Trello, YouTrack</li>
```
cstryker, line 10: since these are listed in alphebetical order (not sure if it matters), maybe move Shortcut to maintain that?

shared/ui/Stream/icons8-data.json
cstryker: It's a close call, but I say let's bite the bullet and do a migration. I'll create a separate card for that.

cstryker (approved):

PR #955, opened 2022-04-26
NR-2200 Rename and add typings to webview .js files 

dsellarsnr:
- Typing Stream/index.js revealed lots of type errors and dead code, sometimes it was missing props passed to other components. For those components, made the props optional since they were never passed in. Probably riskiest part of refactor since it is hard to fully test all the components (Codemark.tsx for example). 
- Typescript 4.5.4 tightened up the type system and revealed lots of type errors which spanned several files. 
- Typescript 4.5.4 hit `styled-components` pretty hard - had to upgrade and do minor refactoring.
- Using Typescript compile option `useUnknownInCatchVariables=false` to defer refactor of all error handling code to a later date 
- Monkey patch in vscode/system/date.ts stopped working in Typescript 4.5.4, using patch-package instead to patch at build time. 

**This PR Addresses:**  
[NR-2200 Rename and add typings to webview .js files ](https://issues.newrelic.com/browse/NR-2200)  


jb/webview/tsconfig.json
```diff
@@ -29,10 +28,11 @@
...
 			"vscode-jsonrpc": ["../../shared/ui/vscode-jsonrpc.shim.ts"]
 		},
-		"rootDir": ".",
+		"skipDefaultLibCheck": true,
 		"skipLibCheck": true,
 		"sourceMap": true,
 		"strict": true,
-		"target": "es2017"
+		"target": "es2017",
+		"useUnknownInCatchVariables": false
```
bcanzanella, line 36: great addition (useUnknownInCatchVariables)

PR #970, opened 2022-05-13
NR-7271 Can't connect to Jira in Ubuntu starting in v12.16 …

dsellarsnr:
- Apply fix to NR agent env vars and merge with existing vars if any
- Fix typing


bcanzanella:
good find.

bcanzanella (approved):

PR #975, opened 2022-05-18
renames some methods

bcanzanella:
renames some methods

### DIFF
--- a/shared/agent/src/providers/provider.ts
+++ b/shared/agent/src/providers/provider.ts
@@ -94,7 +94,7 @@ export interface ThirdPartyProviderSupportsStatus {
 }
 
 export interface ThirdPartyProviderSupportsPullRequests {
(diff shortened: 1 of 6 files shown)


shared/agent/src/providers/provider.ts
```diff
@@ -94,7 +94,7 @@ export interface ThirdPartyProviderSupportsStatus {
 }
 
 export interface ThirdPartyProviderSupportsPullRequests {
-	getRepoInfo(request: ProviderGetRepoInfoRequest): Promise<ProviderGetRepoInfoResponse>;
+	getRepoMetadata(request: ProviderGetRepoInfoRequest): Promise<ProviderGetRepoInfoResponse>;
```
bcanzanella, line 97: Should I also change and request and return types?

Issue #1002, opened 2022-06-06
[VSCode] Trello integration fails


nmay231:
With the latest version of the vscode extension (13.0.0), the Trello integration fails. For example, showing issues from Github and Trello in the issues panel hangs indefinitely, but limiting the selection to only Github finishes without an issue. Attempting to add a note to Trello hangs on "Fetching boards..." Downgrading to the previous version (12.18.0) resolves the issue. There are no error toast notifications; it just hangs.

Github + Trello (Hangs)
![Screenshot from 2022-06-06 01-49-17](https://user-images.githubusercontent.com/35386821/172110247-cc52744f-2bbe-4c3d-8843-22a030326a99.png)

Just Github (Loads fine)
![Screenshot from 2022-06-06 01-48-46](https://user-images.githubusercontent.com/35386821/172110246-6138f604-b190-4368-aae6-bbbdd6614f10.png)

12.18.0 Github + Trello (Loads fine)
![Screenshot from 2022-06-06 01-51-20](https://user-images.githubusercontent.com/35386821/172110525-a2d823a0-e79b-4340-b421-6f1ee3f81877.png)

OS: Ubuntu 20.04
IDE: VSCode 1.67.2 (Also tested with VSCode Insiders1.68.0 with only the Code Stream extension installed)
CodeStream: 13.0.0

PR #1006, opened 2022-06-06
NR-21354: fix trello and asana, timing issues on connection

cstryker:
all this provider stuff needs a serious refactor, but here is another hack on top of all the other hacks

**This PR Addresses:**  
[NR-21354 Trello and Asana integrations appear to be broken (GH #1002)](https://issues.newrelic.com/browse/NR-21354)  


bcanzanella (commented):
looks good.

PR #1011, opened 2022-06-09
NR-22254 - allow users to connect to different GH orgs

cstryker:
Also fixing a display bug with bad camel-case for "GitHub" and "GitLab"

### DIFF
--- a/shared/ui/Stream/ConfigureOAuthOrPATPanel.tsx
+++ b/shared/ui/Stream/ConfigureOAuthOrPATPanel.tsx
@@ -127,7 +127,7 @@ export const ConfigureOAuthOrPATPanel = (props: {
 
 	const connectWithOAuthMsg = derivedState.isVSCGitHub
 		? "Connect with VSCode's Connection"
-		: "Connect With OAuth";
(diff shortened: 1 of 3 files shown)


shared/ui/Stream/ConfigureOAuthOrPATPanel.tsx
```diff
@@ -122,6 +125,9 @@ export const ConfigureOAuthOrPATPanel = (props: {
 		dispatch(closePanel());
 	};
 
+	const connectWithOAuthMsg = derivedState.isVSCGitHub
+		? "Connect with VSCode's Connection"
+		: "Connect With OAuth";
```
bcanzanella, line 130: super nit pick on the differing casings of with & WIth

cstryker: well, i just copied what dave wrote, so, hmmmm.....

bcanzanella: bad dave!

PR #1018, opened 2022-06-15
more verbose codemark creation error for edge cases

eric-j:


shared/ui/store/codemarks/actions.ts
```diff
@@ -91,6 +91,7 @@ export function isLegacyNewCodemarkAttributes(
 
 export interface CreateCodemarkError {
 	reason: "share" | "create";
```
eric-j, line 93: test

PR #1046, opened 2022-07-08
NR-707 - golden metrics in observability panel

eric-j:


shared/agent/src/providers/newrelic.ts
```diff
@@ -2189,7 +2191,7 @@ export class NewRelicProvider extends ThirdPartyIssueProviderBase<CSNewRelicProv
 			return {
 				goldenMetrics: goldenMetrics,
 				newRelicEntityAccounts: observabilityRepo.entityAccounts,
-				newRelicAlertSeverity: entity.alertSeverity,
+				newRelicAlertSeverity: entity.alertSeverity || "hi",
```
bcanzanella, line 2194: "hi"

shared/ui/Stream/Observability.tsx
```diff
@@ -739,111 +735,122 @@ export const Observability = React.memo((props: Props) => {
...
+																				name="link-external"
+																				className="clickable"
+																				title="View on New Relic"
+																				placement="bottomLeft"
+																				delay={1}
+																				onClick={e => {
+																					e.preventDefault();
+																					e.stopPropagation();
+																					HostApi.instance.send(OpenUrlRequestType, {
+																						url: newRelicUrl || " "
```
bcanzanella, line 783: if there's no `newRelicUrl` perhaps this external link icon should be hidden?

bcanzanella (changes requested):
looks good, couple small comments.

eric-j:
@bcanzanella I updated those requested changes

shared/ui/Stream/Observability.tsx
```diff
@@ -394,6 +419,23 @@ export const Observability = React.memo((props: Props) => {
 		}
 	}, [derivedState.hiddenPaneNodes]);
 
+	useInterval(() => {
+		fetchGoldenMetrics(expandedEntity);
+	}, 60000);
```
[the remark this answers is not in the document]
bcanzanella: also, i don't believe i saw anything below, but i think working in some kind of lastUpdated value would be useful for the end users. perhaps it's in a tooltip or somewhere out of the way? That would let the user know that we're getting fresh stats. thoughts?

PR #1086, opened 2022-07-29
VS 2022!

CalvinAllen:
This is a very large PR, so I apologize in advance :)


CalvinAllen:
Because this has been so long running with so many refactors, I didn't do any FRs on it since its basically everything.

vs/build/Extract-Pdb.ps1
```diff
@@ -1,13 +1,13 @@
 #requires -Version 5.0
-$zipPath = "..\src\CodeStream.VisualStudio\bin\x86\Debug\codestream-vs.zip";
+$zipPath = "..\src\CodeStream.VisualStudio.Vsix.x86\bin\x86\Debug\codestream-vs.zip";
 
 if(Test-Path $zipPath) {
 	Remove-Item -Path $zipPath -Force
 }
 
-Copy-Item -Path $(Resolve-Path -Path "..\src\CodeStream.VisualStudio\bin\x86\Debug\codestream-vs.vsix") -Destination "$($(Resolve-Path -Path "..\src\CodeStream.VisualStudio\bin\x86\Debug\"))codestream-vs-CHEESE.vsix"
+Copy-Item -Path $(Resolve-Path -Path "..\src\CodeStream.VisualStudio.Vsix.x86\bin\x86\Debug\codestream-vs.vsix") -Destination "$($(Resolve-Path -Path "..\src\CodeStream.VisualStudio.Vsix.x86\bin\x86\Debug\"))codestream-vs-CHEESE.vsix"
```
bcanzanella, line 8: this might not be used any longer (see name of vsix :/)

vs/build/build.ps1
```diff
@@ -109,58 +111,57 @@ function Build-AgentAndWebview {
...
-		if ($LastExitCode -ne 0) {
-			throw "UnitTests failed"
-		}
-
-		Write-Log "UnitTests completed"
-	}
-	else {
-		Write-Log "UnitTests skipped"
-	}
+	# TODO - how should tests be run when targeting two different VS versions?
```
bcanzanella, line 139: We should get these back running, though i can't recall if we test anything that relies on version-specific VS libraries. 

could we run the current set of tests twice, one for x86 one for x64?

or do we need 2 unit test projects, one for each VS version?

vs/build/Extract-Pdb.ps1
CalvinAllen: I think this does get called from either the pre-build or post-build ps1 files (the ones that are added into the project / VS), maybe that should all be removed then?

vs/build/build.ps1
CalvinAllen: Unfortunately, that's part of my questions, too, and I'm not certain exactly how it should work (forgot to bring it up sooner). It seems like they would definitely need to run in context of version-specific libraries, but I think that would mean two different test suites?

vs/build/Extract-Pdb.ps1
CalvinAllen: Here - https://github.com/TeamCodeStream/codestream/blob/b7c88dac833a7e89a063255c43d36f3597ffede2/vs/src/resources/scripts/Post-Build.ps1#L16

bcanzanella:
🔥 🔥 🔥

vs/build/Extract-Pdb.ps1
bcanzanella: hmm, weird, ok, the 🧀 stuff made me think it was something not used...

vs/build/build.ps1
bcanzanella: I don't think this should hold up getting this in, I'll get a jira in to track it.

Issue #1109, opened 2022-08-16
Observability throughput incorrect


ethanmills:
- IDE Version: IntelliJ IDEA 2022.2.1 RC Ultimate
- OS Version: Windows 10

Displayed throughput in plugin
![image](https://user-images.githubusercontent.com/40488007/184839127-395ce62a-d601-4395-97b1-2ea9d9ae29fc.png)
Throughput in NR
![image](https://user-images.githubusercontent.com/40488007/184839206-5ddead3b-2be7-4674-b388-7b6361575d99.png)

Steps to Reproduce:

1. Link service using plugin
2. Observe

ethanmills:
Looking closer, response time is also incorrect, but closer (122ms vs 89ms)

planteater:
@ethanmills - Could you share a permalink to the service in question on New Relic, or even just the name of the service? If you’d prefer you can email that info to me at support@codestream.com.

ethanmills:
@planteater https://onenr.io/0MR2Ao4oYwY

bcanzanella:
hey @ethanmills could you try the latest version of the CodeStream plugin (v13.5)?

ethanmills:
@bcanzanella Looks great, thanks! Really impressed by how response you folks have been and how quickly this was fixed

PR #1115, opened 2022-08-17
GH #1109 Observability throughput incorrect

bcanzanella:
a lot of these changes are in typings, but one thing I tried to re-work is using "name" as a query key, rather than "title" (from the agent). Any calling code should find queries/queryData using "name" and set the title themselves (as we do with `goldenMetricTitleMapping`)

Also, we're changing the queries we use for service level golden metrics to use the queries used in the NR1 UI.


eric-j (approved):
👍 looks good!

marcelofarias (approved):

PR #1121, opened 2022-08-23
related services in o11y panel

eric-j:
https://issues.newrelic.com/browse/NR-32340
https://issues.newrelic.com/browse/NR-32352
https://issues.newrelic.com/browse/NR-33924
https://issues.newrelic.com/browse/NR-32344


shared/agent/src/providers/newrelic.ts
```diff
@@ -1062,6 +1068,82 @@ export class NewRelicProvider extends ThirdPartyIssueProviderBase<CSNewRelicProv
...
+					  }
+					}
+				  }				  
+			  	`,
+				{
+					entityGuid: request.entityGuid
+				}
+			);
+			if (response?.actor?.entity?.relatedEntities?.results) {
+				const _results = response.actor.entity.relatedEntities.results.map((_: any) => {
```
dsellarsnr, line 1118: We should type this out with an interface instead of using `any`

shared/agent/src/protocol/agent.protocol.providers.ts
```diff
@@ -877,6 +876,13 @@ export interface GetNewRelicErrorGroupRequest {
 	timestamp?: number;
 }
 
+export interface GetNewRelicRelatedEntitiesRequest {
+	entityGuid: string;
+}
+export interface GetNewRelicUrlRequest {
```
eric-j, line 882: test

eric-j:
@dsellarsnr @bcanzanella added better typing to this pr

shared/ui/Stream/ObservabilityGoldenMetricDropdown.tsx
```diff
@@ -129,89 +221,8 @@ export const ObservabilityGoldenMetricDropdown = React.memo((props: Props) => {
...
-												<>No Data</>
-											)}
-										</StyledMetric>
-									</Tooltip>
-								</div>
-							</Row>
-						);
-					})}
-				</>
+			{(noDropdown || expanded) && !loadingGoldenMetrics && !_isEmpty(goldenMetrics) && (
```
dsellarsnr, line 224: Nice, i need to start using isEmpty

dsellarsnr (approved):

shared/agent/src/providers/newrelic.ts
```diff
@@ -2226,29 +2306,32 @@ export class NewRelicProvider extends ThirdPartyIssueProviderBase<CSNewRelicProv
...
-				newRelicEntityAccounts: observabilityRepo.entityAccounts,
-				newRelicAlertSeverity: entity.alertSeverity,
-				newRelicEntityName: entity.entityName!,
-				newRelicEntityGuid: entity.entityGuid!,
-				newRelicUrl: `${this.productUrl}/redirect/entity/${entity.entityGuid}`
+				newRelicEntityAccounts: observabilityRepo?.entityAccounts || [],
+				newRelicAlertSeverity: entity?.alertSeverity,
+				newRelicEntityName: entity?.entityName!,
+				newRelicEntityGuid: entity?.entityGuid! || request.newRelicEntityGuid,
+				newRelicUrl: `${this.productUrl}/redirect/entity/${entity?.entityGuid ||
```
bcanzanella, line 2333: could this call `this.getNewRelicUrl()` ?

shared/ui/Stream/ObservabilityRelatedWrapper.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export const ObservabilityRelatedWrapper = React.memo((props: Props) => {
+	const [expanded, setExpanded] = useState<boolean>(true);
+	const [relatedEntities, setRelatedEntities] = useState<GetNewRelicRelatedEntitiesResponse>([]);
+	const [loadingRelatedEntities, setLoadingRelatedEntities] = useState<boolean>(true);
+
+	useDidMount(() => {
+		(async () => {
+			setLoadingRelatedEntities(true);
```
bcanzanella, line 24: we use this pattern a lot in our react code... set loading, call agent, set response, stop loading. nothing to change here, but perhaps something we could wrap up in another component, or maybe that would be too over-engineered? @eric-j @dsellarsnr

bcanzanella (approved):

shared/ui/Stream/ObservabilityRelatedWrapper.tsx
eric-j: That's a really good point.  I am making some adjustments to this PR and tinkering with a custom hook for this currently.

shared/ui/Stream/ObservabilityRelatedWrapper.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const ObservabilityRelatedWrapper = React.memo((props: Props) => {
+	const [expanded, setExpanded] = useState<boolean>(true);
+	const {
+		loading,
+		data
+	}: {
+		loading: boolean;
+		data: GetNewRelicRelatedEntitiesResponse;
+	} = useRequestType(GetNewRelicRelatedEntitiesRequestType, { entityGuid: props.entityGuid });
```
eric-j, line 25: @bcanzanella here the custom hook I made to abstract that commonly used code of:
setLoading false,
await response
set response
setLoading false

Something interesting to move forward with.  Because lots of older code (see Observabilty.tsx useDidMount) has many nested calls with crazy setStates, its not applicable to everything.  But with new development and simpler smaller components I think it makes a lot of sense.

shared/ui/utilities/hooks.ts
```diff
@@ -49,6 +50,30 @@ export function useInterval(callback: Fn, delay = 1000) {
 	}, [delay]);
 }
 
+/**
+ * @param requestType
+ * @param payload
+ * @param dependencies
+ * @returns [loading, data]
+ */
+export function useRequestType(requestType, payload = {}, dependencies = []) {
```
eric-j, line 59: new custom hook to reduce useEffect await graphql boilerplate calls

shared/agent/src/providers/newrelic.ts
```diff
@@ -45,6 +45,12 @@ import {
 	GetNewRelicErrorGroupRequest,
 	GetNewRelicErrorGroupRequestType,
 	GetNewRelicErrorGroupResponse,
+	GetNewRelicRelatedEntitiesRequest,
+	GetNewRelicRelatedEntitiesRequestType,
+	GetNewRelicRelatedEntitiesResponse,
+	GetNewRelicUrlRequest,
```
eric-j, line 51: test

shared/ui/Stream/ObservabilityRelatedWrapper.tsx
bcanzanella: i dig it!!!! thoughts about error handling? try/catch or something that bubbles up to parent component? 

also, to take the TypeScript part of it a step further, it should be possible to use generics to not require the `{loading, data}` type as well as type-checking the request payload. while the following does make it a little more wordy (many of our Request/Response agent types are so long), something like this might work....


```
interface UseRequestTypeResult<T> {
	data: T | undefined;
	loading: boolean;
}
/**
 * @param requestType<Req, Resp>
 * @param payload
 * @param dependencies
 * @returns [loading, data]
 */
export function useRequestType<Req, Resp>(
	requestType: RequestType<Req, Resp, void, void>,
	payload: Req,
	dependencies = []
): UseRequestTypeResult<Resp> {
	const [loading, setLoading] = useState(true);
	const [data, setData] = useState<Resp | undefined>(undefined);

	const fetch = async () => {
		setLoading(true);
		const response = (await HostApi.instance.send(requestType, payload)) as Resp;
		setData(response);
		setLoading(false);
	};

	useEffect(() => {
		fetch();
	}, dependencies);

	return { loading, data } as UseRequestTypeResult<Resp>;
}
```

... and it would make the usage look like this:

```
	const { loading, data } = useRequestType<
		GetNewRelicRelatedEntitiesRequest,
		GetNewRelicRelatedEntitiesResponse
	>(GetNewRelicRelatedEntitiesRequestType, { entityGuid: props.entityGuid, removeThis: true });
```

`data` would have the correct type and note the TS error with `removeThis` -- since it's not on the `GetNewRelicRelatedEntitiesRequest`

bcanzanella: (there might be a way to only use the `*RequestType` as the generic, but i'd have dig in a big more...)

PR #1139, opened 2022-08-30
Lock PR fails if user doesn't select a reason

reneelpetit:


bcanzanella:
not sure what happened here, but this wasn't ready to be merged in :/

PR #1150, opened 2022-09-02
NR-42458 Upgrade tslint to eslint and introduce commit hooks and lint-staged …

dsellarsnr:
To run checks on git staged files (the same command used in commit hook):
`npx lint-staged` in shared/ui or shared/agent

To run checks on all files:
`npm run verify` in shared/ui or shared/agent

Note: `npm run verify` runs checks in parallel and aborts as soon as any of the 3 verifications fail - to run individually

`npm run verify:compile`
`npm run verify:lint`
`npm run verify:prettier`


.husky/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env sh
```
CalvinAllen, line 1: Since this is at the root of the repo, are these going to run whether we have changes to agent/ui or not?

dsellarsnr: Yes, unfortunately. Let me see if I can improve it.

CalvinAllen: I'm also considering how this will play when I add a similar thing for the VS side, and how they can all play nicely together without taking up too much time

dsellarsnr: Removing tests for now - too slow for a commit hook. But still could do something smart about the `npm run verify:compile` which only needs to be run when certain files are modified.

CalvinAllen: I've done this in the past for C# files (and will likely again for the VS side of the house at some point -

```
FILES=$(git diff --cached --name-only --diff-filter=ACM "*.cs" "*.cshtml")

[ -z "$FILES" ] && exit 0
```

bcanzanella (approved):

PR #1151, opened 2022-09-06
Sorting of commits in PRs doesn't take year into account

reneelpetit:

### DIFF
--- a/shared/ui/Stream/PullRequestCommitsTab.tsx
+++ b/shared/ui/Stream/PullRequestCommitsTab.tsx
@@ -137,10 +137,10 @@ export const PullRequestCommitsTab = props => {
 
 	const [isLoading, setIsLoading] = useState(true);
 	const [commitsByDay, setCommitsByDay] = useState<{
-		[Identifier: string]: FetchThirdPartyPullRequestCommitsResponse[];
+		[identifier: string]: FetchThirdPartyPullRequestCommitsResponse[];
 	}>({});
 
-	const _mapData = data => {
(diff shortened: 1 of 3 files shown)


shared/ui/Stream/PullRequestCommitsTab.tsx
```diff
@@ -132,20 +136,20 @@ export const PullRequestCommitsTab = props => {
 	});
 
 	const [isLoading, setIsLoading] = useState(true);
-	const [commits, setCommits] = useState<any>({});
+	const [commitsByDay, setCommitsByDay] = useState<{
+		[Identifier: string]: FetchThirdPartyPullRequestCommitsResponse[];
```
dsellarsnr, line 140: Very minor javascript convention thing - Identifier should be lowercase

shared/ui/Stream/PullRequestCommitsTab.tsx
```diff
@@ -132,20 +136,20 @@ export const PullRequestCommitsTab = props => {
...
+	}>({});
 
 	const _mapData = data => {
```
dsellarsnr, line 143: You can make this typed - `const _mapData = (data: FetchThirdPartyPullRequestCommitsResponse[]) => {`

A couple of the _mapData calls that come from dispatch won't like that - you'll can fix those with `_mapData(data as any);` for now. There is a better fix coming for the dispatch calls soon.

dsellarsnr (approved):
Nice!

PR #1159, opened 2022-09-09
NR-42250 DEBT: pay down TypeScript `any` usage webview

dsellarsnr:
- Use redux-toolkit to help with typing thunks
- Big step forward but does not close out use of `any`


vscode/src/api/tokenManager.ts
```diff
@@ -3,7 +3,7 @@ import { Logger } from "logger";
 import { GlobalState } from "../common";
 import { extensionId } from "../constants";
 import { Container } from "../container";
-import { keychain } from "../keychain";
+import keychain from "keytar";
```
bcanzanella, line 6: does this bypass that getNodeModule code?

bcanzanella: ahh i see it in the externals... maybe a comment here to mention that?

shared/ui/store/streams/reducer.ts
```diff
@@ -64,8 +64,8 @@ export const getChannelStreamsForTeam = createSelector(
 	(state: CodeStreamState) => state.streams,
 	(_, teamId: string) => teamId,
 	(state: CodeStreamState) => state.session.userId!,
-	(state, teamId, userId) => {
-		const streams = state.byTeam[teamId];
+	(streamState, teamId, userId) => {
+		const streams: CSStream[] = streamState.byTeam[teamId]; // TODO why is this not auto-typed
```
bcanzanella, line 68: hmm, maybe that empty `{}` on line 13?

vscode/src/api/tokenManager.ts
dsellarsnr: Yes, it was the cause of the scary looking eval warning in esbuild. Will add comment.

shared/ui/Stream/OpenPullRequests.tsx
```diff
@@ -262,8 +262,8 @@ export const OpenPullRequests = React.memo((props: Props) => {
 			context.currentPullRequest.id;
 		const currentPullRequest = getCurrentProviderPullRequest(state);
 		const expandedPullRequestGroupIndex = context.currentPullRequest?.groupIndex;
-		const panePreferences = preferences.sidebarPanes || EMPTY_HASH;
-		const settings = panePreferences["open-pull-requests"] || EMPTY_HASH;
+		const panePreferences = preferences.sidebarPanes;
```
bcanzanella, line 265: we ok not having these `EMPTY_HASH*`es?

shared/ui/Stream/PullRequest.tsx
```diff
@@ -245,18 +241,14 @@ export const PullRequest = () => {
...
-			getPullRequestConversations(
-				derivedState.currentPullRequestProviderId!,
-				derivedState.currentPullRequestId!
-			)
-		)) as {
-			error?: {
-				message: string;
-			};
-		};
+		const response = await dispatch(
```
bcanzanella, line 244: nothing to do here. just thinking out loud. would there be any benefit to having a `typedDispatch` (or some other name) that runs an action and calls unwrap() on its own? is there overhead to calling unwrap()?

shared/ui/Stream/PullRequestFilesChangedFileComments.tsx
```diff
@@ -136,7 +133,7 @@ export const PullRequestFilesChangedFileComments = (props: Props) => {
 
 	const syncCheckedStatusWithPr = () => {
 		if (currentPr && !isGitLab && supportsViewerViewedState) {
-			const prFiles = currentPr.files.nodes;
+			const prFiles = (currentPr as any).files.nodes; // TODO Fix typing - does this crash on gitlab since it files is only on github?
```
bcanzanella, line 136: i think the !isGitLab above guards that

shared/ui/Stream/OpenPullRequests.tsx
dsellarsnr: Ahh good catch.  I was getting rid of using {} as a type which is a eslint warning but it looks like it is needed here. Just not a fan of EMPTY_HASH - is {} ok?

shared/ui/Stream/PullRequest.tsx
dsellarsnr: Will look into it

shared/ui/store/streams/reducer.ts
dsellarsnr: The whole thing isn't typed - streamState is `any`. It should infer it from the selector return type. Managed to manually type it.

bcanzanella (approved):
🚀

PR #1162, opened 2022-09-09
Don't double post to slack when messages can't be edited

purplexa:


shared/ui/Stream/actions.ts
```diff
@@ -561,6 +561,7 @@ export const editPost = (
 						}
 					);
 				} catch (error) {
+					if (error.includes("edit_window_closed")) continue;
```
bcanzanella, line 564: is this assuming too much? will `error` _always_ be a string? perhaps check for error.message as well?

bcanzanella: hmm, i guess if it's just a toString of `new Error()` should be ok.

bcanzanella (approved):

PR #1204, opened 2022-10-05
CI/CD fixes

purplexa:


shared/ui/Stream/CICD/CICD.tsx
```diff
@@ -54,8 +54,9 @@ export const CICD = (props: Props) => {
 			providerInfo,
 			providers,
 			currentRepo,
+			currentBranch: editorContext.scmInfo?.scm?.branch,
 		};
-	});
+	}, shallowEqual);
```
dsellarsnr, line 59: Nice!

dsellarsnr (approved):

shared/ui/Stream/CICD/CircleCIBuilds.tsx
```diff
@@ -32,7 +32,7 @@ export const CircleCIBuilds = (props: Props) => {
 						></PaneNodeName>
 						<div style={{ padding: "0 20px 0 40px" }}>
 							{!projectsCollapsed[name] &&
-								workflows.map(workflow => {
+								workflows.slice(0, 5).map(workflow => {
```
bcanzanella, line 35: could/should this be limited at the agent level? are we fetching more data than we need for the UI just to ignore it?

purplexa:
@bcanzanella - okay, we still fetch more data than necessary from CircleCI because I can't find a way to limit the number of results, but we're not making unnecessary requests anymore like we were before.

bcanzanella (approved):

PR #1214, opened 2022-10-12
Ensure all setUserPreference calls use correct arguments

purplexa:
Among other things, this fixes the broken "Auto-prompt for feedback when committing" toggle on the review form.


dsellarsnr:
Oh man what a find! Can you also fix the root cause - the lack of typing on the prop. `Function` is basically `any` Should be a compile error. Looks like this should work:

```
import { setUserPreference } from "./actions";
...
setUserPreference?: typeof setUserPreference;
```

dsellarsnr (approved):
Awesome!

PR #1220, opened 2022-10-14
Fix issue with solution file as path vs. uri for CLM in VS

CalvinAllen:


vs/src/CodeStream.VisualStudio.Shared/Services/CodeLevelMetricsCallbackService.cs
```diff
@@ -80,7 +80,7 @@ public class CodeLevelMetricsCallbackService : ICodeLensCallbackListener, ICodeL
 			}
 		
 			await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-			var solution = _vsSolution.GetSolutionFile();
+			var solution = new Uri(_vsSolution.GetSolutionFile());
```
CalvinAllen, line 83: Another QA test from lightbulb

vs/src/CodeStream.VisualStudio.Shared/Services/CodeLevelMetricsCallbackService.cs
```diff
@@ -90,7 +90,7 @@ public class CodeLevelMetricsCallbackService : ICodeLensCallbackListener, ICodeL
 
 			try {
 				var metrics = await _codeStreamAgentService.GetFileLevelTelemetryAsync(
-					solution,
+					solution.AbsoluteUri,
 					"csharp",
 					false,
 					codeNamespace,
```
CalvinAllen, line 96: Use AbsoluteUri to get the file:// uri version

vs/src/CodeStream.VisualStudio.Shared/Services/CodeLevelMetricsCallbackService.cs
```diff
@@ -90,7 +90,7 @@ public class CodeLevelMetricsCallbackService : ICodeLensCallbackListener, ICodeL
 
 			try {
 				var metrics = await _codeStreamAgentService.GetFileLevelTelemetryAsync(
-					solution,
+					solution.AbsoluteUri,
 					"csharp",
 					false,
```
CalvinAllen, line 95: Test

vs/src/CodeStream.VisualStudio.Shared/Services/CodeLevelMetricsCallbackService.cs
CalvinAllen, line 83: PR comment

PR #1235, opened 2022-10-27
NR-60240 Determine what the minimum version of GHE that codestream is compatible with …

dsellarsnr:
set ghe min version for PRs to 3.3 and add min version warning to ghe integration connect screen

**This PR Addresses:**  
[NR-60240 Determine what the minimum version of GHE that codestream is compatible with](https://issues.newrelic.com/browse/NR-60240)  


bcanzanella (approved):

dsellarsnr:
@bcanzanella added tooltip instead of external link per Dave

PR #1270, opened 2022-11-17
Golden Metrics for all Entity Types

CalvinAllen:

### DIFF
--- a/shared/ui/Stream/Observability.tsx
+++ b/shared/ui/Stream/Observability.tsx
@@ -608,11 +608,11 @@ export const Observability = React.memo((props: Props) => {
 				repoId: currentRepoId,
 				fetchRecentAlertViolations: true,
 			});
-			if (response?.entityGoldenMetrics) {
-				setEntityGoldenMetrics(response.entityGoldenMetrics);
-				setRecentAlertViolations(response.recentAlertViolations);
-				setNewRelicUrl(response.newRelicUrl);
-			}
+
+			setEntityGoldenMetrics(response.entityGoldenMetrics);
+			setRecentAlertViolations(response.recentAlertViolations);
+			setNewRelicUrl(response.newRelicUrl);
+
 			setLoadingGoldenMetrics(false);
 		}
 	};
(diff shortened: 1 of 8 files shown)


shared/ui/Stream/Observability.tsx
```diff
@@ -1132,7 +1127,7 @@ export const Observability = React.memo((props: Props) => {
 																									/>
 																								</>
 																							}
-																							{
+																							{ea.domain === "APM" && (
```
bcanzanella, line 1130: we had assumed here that we'd only see APM entity errors right? for non APM does the UI look ok?

shared/ui/Stream/ObservabilityGoldenMetricDropdown.tsx
```diff
@@ -1,141 +1,52 @@
...
-						let roundToValue = logValue > 2 ? logValue : 2;
-						goldenMetricValue = Number(goldenMetricValue)?.toFixed(roundToValue);
-						noCommas = true;
-					}
-					// add commas to numbers
-					if (goldenMetricValue && !noCommas) {
-						goldenMetricValue = goldenMetricValue.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
-					}
+				{entityGoldenMetrics?.metrics.map(gm => {
+					const goldenMetricDisplayUnit = unitMappings[gm?.unit];
```
bcanzanella, line 49: 👍🏻

bcanzanella: thumbsup to the reduction in code... but would it make sense to store that unitMappings in the agent, and augment the data that is returned to the webview with each metric already having the correctly mapped unit?

CalvinAllen: Yeah, I was actually considering that earlier today, I think it would make reuse a little easier in the future.

shared/ui/Stream/Observability.tsx
CalvinAllen: In this case the entire "Errors" node is gone - looks okay to me, demo forthcoming in standup :)

PR #1287, opened 2022-12-16
NR-72973 Wrong service expanded by default after reload.

dsellarsnr:
**This PR Addresses:**  
[NR-72973 Wrong service expanded by default after reload.](https://issues.newrelic.com/browse/NR-72973)  

### DIFF
--- a/shared/ui/Stream/Observability.tsx
+++ b/shared/ui/Stream/Observability.tsx
@@ -77,50 +77,64 @@ interface Props {
 
 const Root = styled.div`
 	height: 100%;
+
 	.pr-row {
 		padding-left: 40px;
+
 		.selected-icon {
 			left: 20px;
 		}
 	}
+
 	${PaneNode} ${PaneNode} {
 		${PaneNodeName} {
 			padding-left: 40px;
 		}
+
 		.pr-row {
 			padding-left: 60px;
+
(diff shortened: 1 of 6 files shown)


shared/ui/Stream/Observability.tsx
```diff
@@ -326,50 +341,33 @@ export const Observability = React.memo((props: Props) => {
...
+		await getObservabilityErrors();
+	};
+
+	const getObservabilityErrors = async () => {
+		if (currentRepoId) {
+			loading(currentRepoId, true);
 			try {
 				const response = await HostApi.instance.send(GetObservabilityErrorsRequestType, {
-					filters: buildFilters(repoIds),
+					filters: buildFilters([currentRepoId]),
```
bcanzanella, line 366: i'm sure wrote this, but why in the world were we passing all the repos here?? your change is probably a sizable perf gain

bcanzanella (approved):
👍🏻

shared/ui/Stream/Observability.tsx
dsellarsnr: I think there was a time where you could have multiple entities expanded, much of the code was designed for that. I was hoping for a perf boost, but it didn't seem to help much.

PR #1294, opened 2022-12-21
'Missing Git Info' popup has text rendered outside of the popup

bcanzanella:


dsellarsnr (commented):
Looks good but there is probably a guideline somewhere that you shouldn't have components in actions.

dsellarsnr (approved):

bcanzanella:
> Looks good but there is probably a guideline somewhere that you shouldn't have components in actions.

that is a very good call...

PR #1296, opened 2022-12-22
Error in Create step of instrumentation wizard

bcanzanella:
Handle when code isn't a git repo... also adds fix for PATH issue when running npm install


bcanzanella:
ok, this is ready for review now. thanks.

dsellarsnr (approved):

PR #1299, opened 2023-01-03
Feature/common utils rebase

dsellarsnr:
* No more symlinks, shared code resolved via relative paths
* Eliminate some of the duplicate code - shared date. iterable, searchTree, string now resolved via relative path and path aliases. 
* Add shell script utility to root (more robustness and documentation coming later)


shared/agent/src/agent.ts
```diff
@@ -1,10 +1,11 @@
 "use strict";
 // keep this as the first import
 // tslint:disable:ordered-imports
-import * as NewRelic from "newrelic";
 
 import * as fs from "fs";
 import * as os from "os";
+
+import * as NewRelic from "newrelic";
```
bcanzanella, line 8: can this line stay as the first import? IIRC the NR agent requires that.

shared/util/jest.global.js
```diff
@@ -0,0 +1,3 @@
+module.exports = async function (globalConfig, projectConfig) {
+	process.env.NEW_RELIC_ENABLED = "false";
```
bcanzanella, line 2: nice

bcanzanella (changes requested):
1 comment w/r/t imports

shared/agent/src/agent.ts
dsellarsnr: fixed

bcanzanella (approved):

PR #1310, opened 2023-01-17
Removes option for expand/collapse & removes chevron

reneelpetit:
Removes option for expand/collapse of files with comments & removes chevron.
https://issues.newrelic.com/browse/NR-37196


dsellarsnr (approved):
Looks good!

PR #1329, opened 2023-01-27
polling improvements for maintenance mode

eric-j:


cstryker (commented):
this all looks good, but wasn't there part of this that was supposed to stop polling if the extension is not showing in the sidebar, and resume polling once it is?

eric-j:
@cstryker Thats correct, its not included in this PR.  I spent some time trying to get it to work, but didn't find a working solution.  I can spend more time on it, but I thought it would be worth it to get this code out.  

This code will poll every 10 minutes when not in maintenance mode, resetting the timer when the user hits the api (for whatever reason), and every 1 minute while in maintenance mode.  Works if logged in or logged out.

planteater:
@cstryker @eric-j - While we're in maint mode, I think polling even while CS isn't in view is probably fine. Under the assumption we'd never be in maint mode for all that long.

However, polling to see if we should be in maint mode while CS isn't in view seems more problematic, even with a 10-min interval. We have a TON of users who are essentially inactive. Don't use CS, but still have it installed. Although I guess we don't know how many of those are signed out. In any case, something about that gives me a little concern. 

Can we only poll while CS isn't in view if we're already in maint mode?

eric-j:
@planteater trouble is finding an ide event that can accurately give us the information of whether or not CS is active - I'll poke around in the vscode docs and see if there is something we can use.

We do have an event that works when the webview is focused, but in my testing its not reliable enough/doesn't work if the user swaps between extensions

planteater:
Sounds like this is something that could vary by IDE then. Maybe hit up Marcelo re: JB. If we can find a way to do it definitively in certain IDEs but not others, I'd be fine just not polling in those where we can't tell when the user switches to/from CS.

cstryker (approved):

shared/agent/src/api/apiProvider.ts
```diff
@@ -563,6 +563,8 @@ export interface ApiProvider {
 	verifyConnectivity(): Promise<VerifyConnectivityResponse>;
 	setServerUrl(url: string): void;
 
+	pollForMaintenanceMode(): Promise<any>;
```
bcanzanella, line 566: add a type here

shared/agent/src/api/codestream/codestreamApi.ts
```diff
@@ -2900,4 +2901,39 @@ export class CodeStreamApiProvider implements ApiProvider {
...
+		const controller = new AbortController();
+		const timeout = setTimeout(() => controller.abort(), 5000);
+		const response: PollForMaintenanceModeResponse = {
+			ok: true,
+		};
+
+		try {
+			Logger.log("Verifying API server connectivity");
+
+			const resp = await customFetch(this.baseUrl + "/no-auth/capabilities", {
```
bcanzanella, line 2915: can `resp` be consolidated with `response` ?

shared/agent/src/session.ts
```diff
@@ -928,6 +946,13 @@ export class CodeStreamSession {
 		return response;
 	}
 
+	@log({ singleLine: true })
+	async pollForMaintenanceMode(): Promise<PollForMaintenanceModeResponse> {
+		if (!this._api) throw new Error("cannot verify connectivity, no API connection established");
```
bcanzanella, line 951: how does this `throw` bubble up to the user? is it handled well or can it cause any kind of crash?

shared/ui/index.tsx
```diff
@@ -148,7 +151,12 @@ export async function initialize(selector: string) {
 
 	// verify we can connect to the server, if successful, as a side effect,
 	// we get the api server's capabilities and our environment
-	const resp = await HostApi.instance.send(VerifyConnectivityRequestType, void {});
+	const resp: any = await HostApi.instance.send(VerifyConnectivityRequestType, void {});
```
bcanzanella, line 154: is there a typing for this response?

shared/agent/src/session.ts
eric-j: after forcing the error, here is the result:
image

Nothing to note from a user perspective, but console errors are there.

PR #1335, opened 2023-02-03
Watch for new git repos and update displays where appropriate

purplexa:


shared/agent/src/git/repositories.ts
```diff
@@ -488,6 +489,37 @@ export class GitRepositories {
...
 
+		// This watches for when a user clones/copies/creates/etc a new repository in
+		// an existing workspace folder so that we can track the repository properly.
+		const workspaceFolders = await this.session.getWorkspaceFolders();
+		if (workspaceFolders && workspaceFolders.length > 0) {
+			const basePaths = workspaceFolders.map(_ => `${url.fileURLToPath(_.uri)}/**/.git`);
+			const newCloneWatcher = chokidar
+				.watch(basePaths, { cwd: "/", ignoreInitial: true })
+				.on("addDir", async gitPath => {
+					const pathElements = gitPath.split("/");
```
bcanzanella, line 500: is gitPath normalized to using forward slashes? will this work on windows paths?

shared/agent/src/git/repositories.ts
```diff
@@ -488,6 +489,37 @@ export class GitRepositories {
 
 		const { git } = SessionContainer.instance();
 
+		// This watches for when a user clones/copies/creates/etc a new repository in
+		// an existing workspace folder so that we can track the repository properly.
+		const workspaceFolders = await this.session.getWorkspaceFolders();
```
dsellarsnr, line 494: i have a feeling this won't work on JB, will test it out later today. The existing file watch stuff already uses workspace folders - but JB's definition of a workspace folder is where you officially associate a framework (usuall java) with a particular folder or add a module. Cloning a repo that is a node project doesn't get the workspace folder designation even though it gets wired up and detected by JB.

dsellarsnr: Ok it is much better for cloning a repo in a project folder - found another edge case in JB though:

- Open a project - let's say it is /proejcts/project1
- Go to File -> Project structure
- Click Modules and click + to add a module
- Choose import module and browse to a project outside the main project root - for example /projects/project2
- Click ok and open a file in the repo
- o11y is blank unless you restart the IDE

shared/agent/src/git/repositories.ts
purplexa: according to the chokidar documentation, you have to use forward slashes even on windows. I believe the `cwd: "/"` also works based on what I read, but I didn't actually test on windows so it's possible I'm misunderstanding the docs.

shared/agent/src/git/repositories.ts
```diff
@@ -488,6 +489,37 @@ export class GitRepositories {
...
+		if (workspaceFolders && workspaceFolders.length > 0) {
+			const basePaths = workspaceFolders.map(_ => `${url.fileURLToPath(_.uri)}/**/.git`);
+			const newCloneWatcher = chokidar
+				.watch(basePaths, { cwd: "/", ignoreInitial: true })
+				.on("addDir", async gitPath => {
+					const pathElements = gitPath.split("/");
+					if (pathElements[pathElements.length - 1] === ".git") {
+						const name = pathElements[pathElements.length - 2];
+						const uri = url.pathToFileURL(`/${pathElements.slice(0, -1).join("/")}`);
+						await this.onWorkspaceFoldersChanged({
```
bcanzanella, line 504: maybe add a Log line of the name + uri?

shared/agent/src/git/repositories.ts
```diff
@@ -488,6 +489,37 @@ export class GitRepositories {
 
 		const { git } = SessionContainer.instance();
 
+		// This watches for when a user clones/copies/creates/etc a new repository in
+		// an existing workspace folder so that we can track the repository properly.
+		const workspaceFolders = await this.session.getWorkspaceFolders();
+		if (workspaceFolders && workspaceFolders.length > 0) {
+			const basePaths = workspaceFolders.map(_ => `${url.fileURLToPath(_.uri)}/**/.git`);
```
bcanzanella, line 496: do node_modules routinely contain .git dirs?

shared/agent/src/git/repositories.ts
bcanzanella: i like that the agent has all the logic for this. @dsellarsnr I'm assuming we'd need some custom JB-level watcher (or event listener) to make this work? If so, perhaps we punt on that edge case.

dsellarsnr: There should be an event for it already firing (didChangeWorkspaceFolders), maybe something isn't wired up all the way. Can definitely punt.

shared/agent/src/git/repositories.ts
bcanzanella: looks like `find . -type d -name '.git'` at our repo root only shows our root .git dir, so we should be good.

bcanzanella (approved):

PR #1336, opened 2023-02-06
NR-84166 Refactor newrelic.ts & NRManager.ts …

dsellarsnr:
Start exploring dependency injection. Most DI frameworks are off the table since esbuild doesn't support emitDecoratorMetadata which is required for all the DI frameworks that use annotations (https://github.com/evanw/esbuild/issues/257)

Explored breaking up codestreamApi by separating the teams related apis to a separate class and factoring out the http parts to apiClient.

### DIFF
--- a/shared/agent/src/session.ts
+++ b/shared/agent/src/session.ts
@@ -119,6 +119,11 @@ import { GitRepository } from "./git/models/repository";
 import { Logger } from "./logger";
 import { log, memoize, registerDecoratedHandlers, registerProviders, Strings } from "./system";
 import { testGroups } from "./testGroups";
+import cacheReset from "./managers/cache/cacheReset";
+import { ApiClient } from "./api/codestream/api/apiClient";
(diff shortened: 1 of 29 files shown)


shared/agent/src/session.ts
```diff
@@ -357,14 +368,28 @@ export class CodeStreamSession {
...
 		Logger.log(`Reject unauthorized: ${this.rejectUnauthorized}`);
-		this._api = new CodeStreamApiProvider(
-			_options.serverUrl?.trim(),
-			this.versionInfo,
-			this._httpAgent || this._httpsAgent,
-			this.rejectUnauthorized
-		);
 
-		this._api.useMiddleware({
+		this.appInjector = createInjector()
```
bcanzanella, line 372: 👍🏻 👍🏻 👍🏻

bcanzanella:
talked this over with Dillon. Since this will require a large rewrite, will close, but will reconsider if we're ever in rewrite mode.

PR #1357, opened 2023-02-24
NR-91693 CodeStream consuming XXX% CPU time

dsellarsnr:


jb/src/main/kotlin/agent/AgentService.kt
```diff
@@ -286,6 +291,28 @@ class AgentService(private val project: Project) : Disposable {
         FileUtils.copyToFile(javaClass.getResourceAsStream("/agent/agent.js.map"), agentJsMap)
         FileUtils.copyToFile(javaClass.getResourceAsStream("/webview/index.js.map"), webJsMap)
 
+        if (platform == Platform.MAC_X64 || platform == Platform.MAC_ARM64) {
```
bcanzanella, line 294: can any part of this fail in a way that would be detrimental to the starting up / using of CS? should it be wrapped in a try/catch?

marcelofarias (approved):

PR #1372, opened 2023-03-07
NR-96075 Allow stacktrace resolution with invalid git ref

dsellarsnr:
- Bonus fix for CLM JB crash


shared/agent/src/managers/NRManager.ts
```diff
@@ -101,6 +101,21 @@ export class NRManager {
 		this._dotNetCore = new DotNetCoreInstrumentation(session);
 	}
 
+	async resolvePathAtRef(resolvedPath: string, ref?: string): Promise<boolean> {
```
dsellarsnr, line 104: @bcanzanella actually i wonder if we need this at all - almost all code paths return true now.

marcelofarias (approved):

PR #1392, opened 2023-03-17
NR-93085: Link to VM when not enabled is taking user to staging

reneelpetit:
link to VM adjust to either staging or one.newrelic.com depending on the user

### DIFF
--- a/shared/agent/src/providers/newrelic/vuln/nrVulnerability.ts
+++ b/shared/agent/src/providers/newrelic/vuln/nrVulnerability.ts
@@ -203,7 +203,9 @@ export class NewRelicVulnerabilitiesProvider extends ThirdPartyProviderBase {
 		const json = JSON.parse(body);
 		if (isNrSecError(json)) {
 			if (json.error === "Forbidden" && json.status === 403) {
-				return new ResponseError(ERROR_VM_NOT_SETUP, "VM possibly not setup");
+				return new ResponseError<{ url: string }>(ERROR_VM_NOT_SETUP, "VM possibly not setup", {
+					url: this.productUrl,
+				});
 			} else {
 				Logger.warn(`Unknown error: ${json.status} ${json.error}: ${json.message}`);
 				return new ResponseError(ERROR_GENERIC_USE_ERROR_MESSAGE, "Unknown error");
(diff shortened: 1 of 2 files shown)


shared/agent/src/providers/newrelic/vuln/nrVulnerability.ts
```diff
@@ -209,14 +215,20 @@ export class NewRelicVulnerabilitiesProvider extends ThirdPartyProviderBase {
...
+			const data = await this.doGetLibraryDetails(request);
+			const { rows = 5 } = request;
+			const libraries = rows === "all" ? data : data.slice(0, rows);
+			return {
+				libraries,
+				recordCount: libraries.length,
+				totalRecords: data.length,
+			};
+		} catch (err) {
+			throw new ResponseError<{ url: string }>(err.code || ERROR_VM_NOT_SETUP, err.message, {
```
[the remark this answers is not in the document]
dsellarsnr: You can updated `handleErrorResponse` instead of adding the try / catch, only the ERROR_VM_NOT_SETUP scenario needs the url data.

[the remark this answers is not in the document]
reneelpetit: OK, I think I got it. I pushed the new changes.

shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -256,7 +256,7 @@ export const SecurityIssuesWrapper = React.memo((props: Props) => {
 
 	const { loading, data, error } = useRequestType<
 		typeof GetLibraryDetailsType,
-		ResponseError<void>
+		ResponseError<{ url: string }>
```
dsellarsnr, line 259: This should probably be <unknown> as different errors will have different payloads in the future

shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -293,23 +293,20 @@ export const SecurityIssuesWrapper = React.memo((props: Props) => {
 	}
 
 	const getErrorDetails = React.useCallback(
-		(error: Error): JSX.Element => {
+		(error: ResponseError<{ url: string }>): JSX.Element => {
```
dsellarsnr, line 296: keep this as plain `Error` - inside of the isResponseError if statement is where we check if it is a ResponseError using a type guard and narrow down the type of error. Inside of that is where you can check if it has a url property.

shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -256,7 +269,7 @@ export const SecurityIssuesWrapper = React.memo((props: Props) => {
 
 	const { loading, data, error } = useRequestType<
 		typeof GetLibraryDetailsType,
-		ResponseError<void>
+		ResponseError<{ void }>
```
dsellarsnr, line 272: void without the curly braces

PR #1434, opened 2023-04-13
adds support for migrating serverUrls

bcanzanella:


dsellarsnr (approved):

CalvinAllen (commented):
You didn't copy the token from using the old url to the new url

bcanzanella:
> You didn't copy the token from using the old url to the new url

want to take a crack at it for me @CalvinAllen ?

CalvinAllen:
@bcanzanella When do you need it? I just stepped away for a bit.

bcanzanella:
> @bcanzanella When do you need it? I just stepped away for a bit.

we'll want to get this out Mon or Tues

CalvinAllen:
@bcanzanella Pushed a change to copy the token in VSC, please review

bcanzanella:
@CalvinAllen looks good -- i added some error handling too

vscode/src/extension.ts
```diff
@@ -87,6 +92,36 @@ export async function activate(context: ExtensionContext) {
...
 	}
+	let updatedServerUrl;
+	for (const key of Object.keys(serverUrlMigrations)) {
+		if (cfg.serverUrl === key) {
+			try {
+				await configuration.updateEffective(
+					configuration.name("serverUrl").value,
+					serverUrlMigrations[key]
+				);
+				updatedServerUrl = true;
```
bcanzanella, line 103: @CalvinAllen should this updateServerUrl go here or at the end of the try block?

if the token stuff fails, we'd still want 		cfg = configuration.get<Config>(); to happen..

bcanzanella: hmm, only if the updateEffective actually worked...

CalvinAllen: Well, I was going for quick / easy, since if any of it fails, all we'll do is re-grab the config - and this code would come out at some point anyway 🤷🏻‍♂️

bcanzanella: gotcha, hmm think i will keep it asis.

PR #1441, opened 2023-04-24
Support errors in CLM details views

purplexa:


shared/ui/Stream/MethodLevelTelemetry/MethodLevelTelemetryPanel.tsx
```diff
@@ -374,6 +394,48 @@ export const MethodLevelTelemetryPanel = () => {
...
+													<ErrorRow
+														key={`observability-error-${index}`}
+														title={_.errorClass}
+														tooltip={_.message}
+														subtle={_.message}
+														alternateSubtleRight={`${_.count}`} // we want to show count instead of timestamp
+														url={_.errorGroupUrl}
+														customPadding={"0"}
+														onClick={async e => {
+															try {
```
eric-j, line 412: Earlier today I merged in some loading changes to places where we make this call and the subsequently dispatch `openErrorGroup`.  This area could use them too.  Something like:

```
setIsLoadingErrorGroupGuid(_.errorGroupGuid);
```

where IsLoadingErrorGroupGuid is a prop passed to the `<ErrorRow />` component.   

It will add a spinner to the beginning of the error row when a user clicks on it making the UX a bit better.  In its current state there will be a significant delay between when a user clicks on the row, and the error modal pops up.


If you pull the latest from `develop`, take a look at where we call this in `ObservabilityErrorDropdown.tsx` for an example.

eric-j (approved):
Nice! One small comment on loading states otherwise 👍

PR #1467, opened 2023-05-10
updating bitbucket.ts so in sync with slice.ts

reneelpetit:
To fix refresh issues, updating bitbucket.ts info to match slice.ts info


eric-j (approved):
👍

PR #1481, opened 2023-05-23
BB & GH - Commit list infinite loading exception NR-115899

reneelpetit:
A PR created in within Codestream, on the details screen, the commit tab, no longer has infinite loading or error.


eric-j (approved):
👍

PR #1482, opened 2023-05-23
a/b test for simplified ui on nr signup

eric-j:


shared/ui/Authentication/SignupNewRelic.tsx
```diff
@@ -127,6 +128,9 @@ export const SignupNewRelic = () => {
 				});
 			};
 
+			const machineIdNumeric = derivedState.machineId.replace(/\D/g, "").substring(0, 10);
```
bcanzanella, line 131: @eric-j  IIRC machineId only exists for VSCode -- is that ok? if not, maybe we use something like the first digit found of the CodeStream userId (% 2)? cc: @dsellarsnr

shared/ui/Authentication/actions.ts
```diff
@@ -293,6 +300,33 @@ export const onLogin =
 			})
 		);
 
+		if (nrSignupTestUi) {
+			await dispatch(
```
bcanzanella, line 304: @eric-j there is a `setUserPreferences` that takes multiple keys, it needs some typescript love on the input parameter (there's an `any` there), but that could turn this into a single update rather than 4.

shared/ui/Authentication/SignupNewRelic.tsx
eric-j: Yeah I can use the userId. I think I will have to move the determining A or B logic further down the flow since we don't have the userID at this point, but I don't think thats an issue.

eric-j:
@bcanzanella I updated setUserPreferences to inline with setUserPreference (note the plural to singular)

shared/ui/Authentication/SignupNewRelic.tsx
bcanzanella: oh hmm... you are right, didn't realize this was in the Signup flow...

PR #1483, opened 2023-05-24
Grok UI

dsellarsnr:
Started out as a branch for selecting code block from the IDE, turned in to full Grok UI. @marcelofarias - requested for JB changes. This is behind a feature flag so safe to merge to develop after release.


vscode/src/controllers/symbolEditController.ts
```diff
@@ -0,0 +1,90 @@
...
+): Promise<EditorCopySymbolResponse> {
+	const editor = await Editor.findOrOpenEditor(Uri.parse(params.uri));
+	if (!editor?.document) {
+		return { success: false };
+	}
+	const symbols = await symbolLocator.locate(editor?.document, new CancellationTokenSource().token);
+	const fullList = [...symbols.allSymbols, ...symbols.instrumentableSymbols.map(s => s.symbol)];
+	for (const symbol of fullList) {
+		// Strip () out of method name .i.e getVets() becomes getVets
+		const simpleSymbolName = symbol.name.replace(/\(.*?\)$/, "");
```
bcanzanella, line 26: Lexa was working on fix for something very close to that with https://github.com/TeamCodeStream/codestream/pull/1480

vscode/src/controllers/symbolEditController.ts
```diff
@@ -0,0 +1,90 @@
...
+		success: false
+	};
+}
+
+export async function replaceSymbol(
+	params: EditorReplaceSymbolRequest
+): Promise<EditorReplaceSymbolResponse> {
+	const uri = Uri.parse(params.uri);
+	const editor = await Editor.findOrOpenEditor(uri);
+	const symbolLocator = new SymbolLocator();
```
bcanzanella, line 47: minor nitpick. this could move to after the editor.document check

vscode/src/controllers/symbolEditController.ts
```diff
@@ -0,0 +1,90 @@
...
+				range: symbol.range
+			};
+		}
+	}
+	return {
+		success: false
+	};
+}
+
+export async function replaceSymbol(
```
bcanzanella, line 42: any idea what happens in the case of an unhandled exception here? maybe wrap in a try/catch?

vscode/src/controllers/symbolEditController.ts
```diff
@@ -0,0 +1,90 @@
...
+	EditorReplaceSymbolResponse
+} from "../../../shared/ui/ipc/host.protocol.editor";
+import { Editor } from "extensions";
+import { Uri, commands, Range } from "vscode";
+import { CancellationTokenSource } from "vscode-languageclient";
+import { BuiltInCommands } from "../constants";
+
+const symbolLocator = new SymbolLocator();
+
+export async function copySymbol(
```
bcanzanella, line 15: any idea what happens in the case of an unhandled exception here? maybe wrap in a try/catch?

shared/ui/Stream/Posts/Reply.tsx
```diff
@@ -199,7 +199,7 @@ const ComposeWrapper = styled.div.attrs(() => ({
 	}
 `;
 
-export const Reply = (props: ReplyProps) => {
+export const Reply = forwardRef((props: ReplyProps, ref: Ref<any>) => {
```
bcanzanella, line 202: for my own education, what does this do?

bcanzanella (approved):
I checked all the non-JB code. couple of suggestions in there, up to you if you think they're worthwhile @dsellarsnr .

shared/ui/Stream/Posts/Reply.tsx
dsellarsnr: By default when you use a custom react component you can't use the `ref` attribute to get access to a HTMLElement - this lets you expose ref and choose which internal dom it will refer to (Root on line 305). The overall goal is to scroll the top of new grok responses to the top of the view.

vscode/src/controllers/symbolEditController.ts
dsellarsnr: 👍  it actually shouldn't be there at all - already created at top of file

shared/ui/Stream/Posts/Reply.tsx
bcanzanella: makes sense, thanks!

jb/src/main/kotlin/agent/AgentService.kt
```diff
@@ -690,10 +690,10 @@ class AgentService(private val project: Project) : Disposable {
         return gson.fromJson(json!!)
     }
 
-    suspend fun getCommit(params: GetCommitParams): GetCommitResult {
+    fun getCommit(params: GetCommitParams): GetCommitResult {
         val json = remoteEndpoint
             .request("codestream/scm/commit", params)
-            .await() as JsonObject?
+            .get() as JsonObject?
```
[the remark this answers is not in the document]
dsellarsnr: Any attempt to use a coroutine results in 

java.lang.Throwable: Read access is allowed from inside read-action (or EDT) only

It basically throws a wrench in the smartInvokeLater.  

I'll mess with it a little more and see if a can get a coroutine that works.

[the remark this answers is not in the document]
marcelofarias: Does ReadAction.nonBlocking(toExecute).submit(NonUrgentExecutor.getInstance()) help?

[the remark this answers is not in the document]
dsellarsnr: ok, check it out now. i found a whole new world of possibilities via withContext

marcelofarias (approved):

PR #1484, opened 2023-05-24
updated changelogs and release versions

kjoshi31:


vs/src/CodeStream.VisualStudio.CodeLens/Properties/AssemblyInfo.cs
```diff
@@ -31,5 +31,5 @@
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("14.20.1.0")]
-[assembly: AssemblyFileVersion("14.20.1.0")]
+[assembly: AssemblyVersion("14.21.0")]
```
bcanzanella, line 34: @kjoshi31 All of the c# versions should use a version with 4 digit places. this should be "14.21.0.0" (same comment for all the other Visual Studio-related changes below). 

Also, I believe there are a bunch of other spots that have versions for VS. Search for "14.20.1.0" to find them

PR #1562, opened 2023-06-29
[FOSSA] Determine Current Repo Context 

angieyeh:
This PR handles part of NR-119852 for determining repo context for display purposes in the section header. The PR does not handle determining which results from FOSSA to display based on a match with the current repo. Working with FOSSA's API will come as part of a future task. 

current-repo-context

### DIFF
--- a/shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
+++ b/shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
@@ -4,11 +4,13 @@ import { ReposScm, FetchThirdPartyCodeAnalyzersRequestType } from "@codestream/p
 import { HostApi } from "@codestream/webview/webview-api";
 import { CodeStreamState } from "@codestream/webview/store";
 import { getUserProviderInfoFromState } from "@codestream/webview/store/providers/utils";
-import { WebviewPanels } from "@codestream/webview/ipc/webview.protocol.common";
-import { PaneBody, PaneHeader, PaneState } from "@codestream/webview/src/components/Pane";
+import { useMemoizedState } from "@codestream/webview/utilities/hooks";
+import { WebviewPanels } from "../../ipc/webview.protocol.common";
+import { PaneBody, PaneHeader, PaneState } from "../../src/components/Pane";
 import Icon from "../Icon";
 import { ConnectFossa } from "./ConnectFossa";
 import { FossaResults } from "./FossaResults";
+import { CurrentRepoContext } from "@codestream/webview/Stream/CurrentRepoContext";
 
 interface Props {
 	openRepos: ReposScm[];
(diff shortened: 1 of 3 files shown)


shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -35,26 +36,14 @@ export const CodeAnalyzers = (props: Props) => {
 			providers,
 		};
 	}, shallowEqual);
+	const [currentRepoId, setCurrentRepoId] = useMemoizedState<string | undefined>(undefined);
```
angieyeh, line 39: I am not sure if CodeAnalyzers needs this state, but I had to add it because CurrentRepoContext requires setCurrentRepoId to be passed as a prop. In CurrentRepoContext, if I set currentRepoCallback as optional I would have to wrap props.currentRepoCallback in an if statement to check that it exists before calling it. 

Using the if statement makes the code not look as clean so I chose to add this state. Let me know what your thoughts are about these two options or if there is a better approach than the ones I've considered!

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -41,17 +43,20 @@ export const CodeAnalyzers = (props: Props) => {
 		};
 	}, shallowEqual);
 
+	const { fossa } = derivedState.providerInfo;
+	const memoizedProviderInfo = useMemo(
+		() => ({
+			fossa,
+		}),
+		[fossa]
+	);
```
angieyeh, line 52: @eric-j regards to moving the loop inside useMemo, I looked into doing that and in line 31, we call `getUserProviderInfoFromState(name, state)` which requires the state object as an arg. Given this is the case, I can't think of a good way to not call the loop in the `useSelector`. thoughts?

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -18,6 +19,7 @@ interface Props {
 export const CodeAnalyzers = (props: Props) => {
 	const [loading, setLoading] = useState(false);
 	const [message, setMessage] = useState("");
+	const [currentRepoId, setCurrentRepoId] = useMemoizedState<string | undefined>(undefined);
```
angieyeh, line 22: I am not sure if CodeAnalyzers needs this state, but I had to add it because the component CurrentRepoContext requires setCurrentRepoId to be passed as a prop. In CurrentRepoContext, if I set the prop currentRepoCallback as optional I would have to wrap props.currentRepoCallback in an if statement to check that it exists before calling it.

Using the if statement makes the code not look as clean so I chose to add this state. Let me know what your thoughts are about these two options or if there is a better approach than the ones I've considered!

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
eric-j: the providers array is never going to be very long, so including the loop within the derivedState declaration should be fine here.

eric-j (approved):

PR #1627, opened 2023-07-19
Clickable Stack Trace Lines

CalvinAllen:
...if the files can be resolved.


CalvinAllen:
Lots of plumbing. The biggest "new thing" is the `FileResolutionService`

vs/src/CodeStream.VisualStudio.Shared/Services/FileResolutionService.cs
```diff
@@ -0,0 +1,70 @@
...
+		private readonly VisualStudioWorkspace _visualStudioWorkspace;
+		//private readonly string _solutionFile;
+
+		[ImportingConstructor]
+		public FileResolutionsService(VisualStudioWorkspace visualStudioWorkspace)
+		{
+			_visualStudioWorkspace = visualStudioWorkspace;
+		}
+
+		public string CanResolve(string relativeFilePath)
```
bcanzanella, line 36: small nitpick on the Name/Args. Since it begins with "Can" it makes be think its truthy/falsey. I would expect something slightly more .NET-like, perhaps `public bool TryCanResolve(string relativeFilePath, out string absolutePath)` or maybe just `public string Resolve(string relativeFilePath)`

bcanzanella: Also, some Exception handling here could be useful. I'm sure if some mangled paths find their way here there could be some authorization-related exceptions.

bcanzanella (approved):

CalvinAllen:
@bcanzanella Renamed method; added error handling - added packages for VS2019

bcanzanella (approved):
LGTM

dsellarsnr (approved):

PR #1631, opened 2023-07-21
Make Recent Errors better match Errors Inbox

purplexa:
Relevant code is on lines 3402 and 3406.

Errors Inbox doesn't group errors based on fingerprint; rather, they use metrics data and heuristics. This attempts to replicate what Errors Inbox as close as possible without major rewrites.


eric-j (approved):
👍

PR #1632, opened 2023-07-22
[FOSSA] Display vulnerabilities and license dependencies

angieyeh:
This PR is for NR-119847. PR contains displaying fossa issues, conditional text for different repo states, and the repo context is also being determined in the agent layer. Error handling in the UI will come in a following PR

Test Cases
- [x] No repo open, no file open
- [x] One repo opened, repo has no remotes, no file opened
- [x] One repo opened, repo has no remotes, open file(s) 
- [x] One repo opened, repo has no remotes, switch between file(s) 
- [x] One repo opened, repo has no fossa match, no file opened
- [x] One repo opened, repo has no fossa match, open file(s) 
- [x] One repo opened, repo has no fossa match, switch between file(s) 
- [x] One repo opened, repo has fossa match, no file opened
- [x] One repo opened, repo has fossa match, open file(s) 
- [x] One repo opened, repo has fossa match, switch between file(s) 
- [x] Multiple repos opened, no file opened
- [x] Multiple repos opened, open file(s) from one repo
- [x] Multiple repos opened, files from some or all repos opened
- [x] Close all files

Edge cases that I ran into:
- CurrentRepoContext doesn't seem to update the section header name (state isn't being updated) when we close out all the files. I checked the staging extension and that happens in the Observability section header as well.
- While testing, in about 1 in 20 times while testing Vuln and Lic Dep, the UI will display the two dropdowns, but the data isn't resolved yet so we display "👍 No X issues found". When the data comes back it will eventually display. A skeleton loader for this issue might be a nice to have. I added a new story for creating two skeleton loaders, one for each fossa issues section (NR-145820)
    - pushed up a fix in CodeAnalyzers fetchData() function in the useEffect
- Repos without remotes are a problem. The way we match the currentRepo to the list of openRepos, we check that the currentRepo id matches an id in openRepos list. A repo without remote(s) has id set to undefined so we will match the currentRepo to the first repo in the openRepos list that has id undefined. If there are multiple opened repos without remotes, we may end up mismatching the currentRepo to the wrong repo in the list. It would also cause the UI to not display repo context switching. 
  - This issue shouldn't be a huge deal since we display a message to users that the git remote doesn't have a remote. Also most people don't have multiple repos opened, let alone multiple without remotes. However since FOSSA allows users to analyze repos without remotes, if we wanted to display the results, it would be challenging with the way we determine the currentRepo. 
   
repo-context-vln-licdep

### DIFF
--- a/shared/agent/src/providers/fossa.ts
+++ b/shared/agent/src/providers/fossa.ts
@@ -79,7 +79,7 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 		category: string,
 		page: number,
 		sort?: string,
-	) {
+	): string {
 		return `/issues?${qs.stringify({
 			category,
 			page,
--- a/shared/agent/src/providers/fossa.ts
+++ b/shared/agent/src/providers/fossa.ts
@@ -72,8 +72,13 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 	 * Query string for Fossa Issues API
 	 * @returns query string
 	 */
-	private _getIssuesUrl(params: any) {
-		const { category, page, sort, projectId, type } = params;
+	private _getIssuesUrl(
+		projectId: string,
+		type: string,
+		category: string,
+		page: number,
+		sort?: string,
+	) {
 		return `/issues?${qs.stringify({
 			category,
 			page,
@@ -189,8 +194,9 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 				return { issues: [] };
 			}
 
+			const { type, category, page, sort } = params;
 			const issueResponse = await this.get<VulnerabilityIssues | LicenseDependencyIssues>(
-				this._getIssuesUrl({ projectId: project.id, ...params }),
+				this._getIssuesUrl(project.id, type, category, page, sort),
 			);
 
 			return {
--- a/shared/agent/src/providers/fossa.ts
+++ b/shared/agent/src/providers/fossa.ts
@@ -4,8 +4,8 @@ import { isEmpty as _isEmpty } from "lodash-es";
 import {
 	FetchThirdPartyCodeAnalyzersRequest,
 	FetchThirdPartyCodeAnalyzersResponse,
-	FetchThirdPartyRepoMatchToFossaProjectRequest,
-	FetchThirdPartyRepoMatchToFossaProjectResponse,
+	FetchThirdPartyRepoMatchToFossaRequest,
+	FetchThirdPartyRepoMatchToFossaResponse,
 	ReposScm,
 	ThirdPartyProviderConfig,
 } from "@codestream/protocols/agent";
@@ -147,19 +147,19 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 	}
 
 	@log()
-	async fetchRepoMatchToFossaProject(
-		request: FetchThirdPartyRepoMatchToFossaProjectRequest,
-	): Promise<FetchThirdPartyRepoMatchToFossaProjectResponse> {
+	async fetchIsRepoMatch(
+		request: FetchThirdPartyRepoMatchToFossaRequest,
+	): Promise<FetchThirdPartyRepoMatchToFossaResponse> {
 		const [currentRepo] = await this.getCurrentRepo(request.repoId);
 		if (!currentRepo) {
-			return { matchedRepoToFossaProject: false };
+			return { isRepoMatch: false };
 		}
 		const projects: FossaProject[] = await this.getProjects();
 		const project = await this.matchRepoToFossaProject(currentRepo, projects, request.repoId);
 		if (_isEmpty(project)) {
-			return { matchedRepoToFossaProject: false };
+			return { isRepoMatch: false };
 		}
-		return { matchedRepoToFossaProject: true };
+		return { isRepoMatch: true };
 	}
 
 	@log()
--- a/shared/agent/src/providers/fossa.ts
+++ b/shared/agent/src/providers/fossa.ts
@@ -55,7 +55,624 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 		request: FetchThirdPartyCodeAnalyzersRequest
 	): Promise<FetchThirdPartyCodeAnalyzersResponse> {
 		return {
-			message: "HELLO WORLD",
+			issues: [
+				{
+					id: 3523803,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+DotNetBrowser32$1.20.0",
+						name: "DotNetBrowser32",
+						url: "https://www.teamdev.com/dotnetbrowser",
+						version: "1.20.0",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523805,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+DotNetBrowser64$1.20.0",
+						name: "DotNetBrowser64",
+						url: "https://www.teamdev.com/dotnetbrowser",
+						version: "1.20.0",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523807,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Microsoft.VisualStudio.CommandBars$8.0.0.1",
+						name: "Microsoft.VisualStudio.CommandBars",
+						url: "https://aka.ms/vsextensibility",
+						version: "8.0.0.1",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523806,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Microsoft.VisualStudio.LanguageServer.Client$16.2.1079",
+						name: "Microsoft.VisualStudio.LanguageServer.Client",
+						url: "https://aka.ms/vslsp",
+						version: "16.2.1079",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523810,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Microsoft.VisualStudio.LanguageServer.Protocol$17.2.8",
+						name: "Microsoft.VisualStudio.LanguageServer.Protocol",
+						url: "https://aka.ms/vslsp",
+						version: "17.2.8",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523804,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Microsoft.VisualStudio.LanguageServer.Protocol$16.2.1079",
+						name: "Microsoft.VisualStudio.LanguageServer.Protocol",
+						url: "https://aka.ms/vslsp",
+						version: "16.2.1079",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523808,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Microsoft.VisualStudio.SDK$16.0.206",
+						name: "Microsoft.VisualStudio.SDK",
+						url: "https://aka.ms/vsextensibility",
+						version: "16.0.206",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523824,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Microsoft.VSSDK.BuildTools$17.2.2186",
+						name: "Microsoft.VSSDK.BuildTools",
+						url: "https://aka.ms/vsextensibility",
+						version: "17.2.2186",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "policy_flag",
+					details:
+						"Requires you to (effectively) disclose your source code if the library is statically linked to your project. Not triggered if dynamically linked or a separate process.",
+					license: "LGPL-3.0-only",
+				},
+				{
+					id: 3523809,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "nuget+Nerdbank.Streams$2.1.37",
+						name: "Nerdbank.Streams",
+						url: "https://github.com/AArnott/Nerdbank.Streams",
+						version: "2.1.37",
+						packageManager: "nuget",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "unlicensed_dependency",
+					license: null,
+				},
+				{
+					id: 3523825,
+					createdAt: "2023-06-23T19:32:05.784Z",
+					source: {
+						id: "npm+newrelic$9.8.1",
+						name: "newrelic",
+						url: "https://www.npmjs.com/package/newrelic",
+						version: "9.8.1",
+						packageManager: "npm",
+					},
+					depths: {
+						direct: 2,
+						deep: 0,
+					},
+					statuses: {
+						active: 2,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream-server",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream-server.git",
+						},
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "policy_flag",
+					details:
+						"Requires you to (effectively) disclose your source code if the library is statically linked to your project. Not triggered if dynamically linked or a separate process.",
+					license: "LGPL-3.0-only",
+				},
+				{
+					id: 3523822,
+					createdAt: "2023-06-23T19:32:05.784Z",
+					source: {
+						id: "npm+newrelic$9.8.1",
+						name: "newrelic",
+						url: "https://www.npmjs.com/package/newrelic",
+						version: "9.8.1",
+						packageManager: "npm",
+					},
+					depths: {
+						direct: 2,
+						deep: 0,
+					},
+					statuses: {
+						active: 2,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream-server",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream-server.git",
+						},
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
+							title: "https://github.com/TeamCodeStream/codestream.git",
+						},
+					],
+					type: "policy_flag",
+					details:
+						"Requires you to (effectively) disclose your source code if the library is statically linked to your project. Not triggered if dynamically linked or a separate process.",
+					license: "LGPL-3.0-or-later",
+				},
+				{
+					id: 3523814,
+					createdAt: "2023-06-23T19:06:36.425Z",
+					source: {
+						id: "npm+newrelic$9.14.1",
+						name: "newrelic",
+						url: "https://www.npmjs.com/package/newrelic",
+						version: "9.14.1",
+						packageManager: "npm",
+					},
+					depths: {
+						direct: 1,
+						deep: 0,
+					},
+					statuses: {
+						active: 1,
+						ignored: 0,
+					},
+					projects: [
+						{
+							id: "custom+38453/github.com/TeamCodeStream/codestream",
+							status: "active",
+							depth: 1,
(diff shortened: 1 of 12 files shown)


shared/agent/src/providers/registry.ts
```diff
@@ -48,12 +48,17 @@ import {
 	FetchThirdPartyChannelsRequestType,
 	FetchThirdPartyChannelsResponse,
 	FetchThirdPartyCodeAnalyzersRequest,
```
angieyeh, line 50: In `fossa.ts` there is one function called fetchCodeAnalysis, but in the `registry.ts` i split fetchCodeAnalysis into `fetchVulnerablities` and `fetchLicenseDependencies` 2 functions. That is why `fetchVulnerablities` and `fetchLicenseDependencies` both use `FetchThirdPartyCodeAnalyzersRequest` and don't have their own.

shared/agent/src/providers/registry.ts
```diff
@@ -48,12 +48,17 @@ import {
...
+	FetchThirdPartyLicenseDependenciesResponse,
 	FetchThirdPartyPullRequestCommitsRequest,
 	FetchThirdPartyPullRequestCommitsType,
 	FetchThirdPartyPullRequestRequest,
 	FetchThirdPartyPullRequestRequestType,
+	FetchThirdPartyRepoMatchToFossaRequest,
+	FetchThirdPartyRepoMatchToFossaRequestType,
+	FetchThirdPartyRepoMatchToFossaResponse,
+	FetchThirdPartyVulnerabilitiesRequestType,
+	FetchThirdPartyVulnerabilitiesResponse,
```
angieyeh, line 61: Change I made

shared/agent/src/providers/registry.ts
```diff
@@ -801,10 +806,10 @@ export class ThirdPartyProviderRegistry {
...
 
 	@log()
-	@lspHandler(FetchThirdPartyCodeAnalyzersRequestType)
-	async fetchCodeAnalysis(
-		request: FetchThirdPartyCodeAnalyzersRequest
-	): Promise<FetchThirdPartyCodeAnalyzersResponse> {
+	@lspHandler(FetchThirdPartyLicenseDependenciesRequestType)
+	async fetchLicenseDependencies(
+		request: FetchThirdPartyCodeAnalyzersRequest,
+	): Promise<FetchThirdPartyLicenseDependenciesResponse> {
```
angieyeh, line 812: Change I made

shared/agent/src/providers/registry.ts
```diff
@@ -819,14 +824,70 @@ export class ThirdPartyProviderRegistry {
...
+		const codeAnalyzersProvider = provider as ThirdPartyCodeAnalyzerProvider;
+		if (
+			codeAnalyzersProvider == null ||
+			typeof codeAnalyzersProvider.supportsCodeAnalysis !== "function" ||
+			!codeAnalyzersProvider.supportsCodeAnalysis()
+		) {
+			throw new Error(`Provider(${provider.name}) doesn't support code analyzers`);
+		}
+
+		const response = await codeAnalyzersProvider.fetchIsRepoMatch(request);
```
angieyeh, line 883: Change I made

shared/agent/src/providers/provider.ts
```diff
@@ -21,6 +21,8 @@ import {
 	FetchThirdPartyCardWorkflowResponse,
 	FetchThirdPartyChannelsRequest,
 	FetchThirdPartyChannelsResponse,
+	FetchThirdPartyRepoMatchToFossaRequest,
+	FetchThirdPartyRepoMatchToFossaResponse,
```
angieyeh, line 25: This file has prettier formatting changes
Change I made

shared/agent/src/providers/provider.ts
```diff
@@ -45,6 +47,8 @@ import { SessionContainer } from "../container";
 import { GitRemote, GitRemoteLike, GitRepository } from "../git/gitService";
 import { Logger } from "../logger";
 
+import { IssueParams } from "../../../util/src/protocol/agent/agent.protocol.fossa";
+
```
angieyeh, line 51: Change I made

shared/agent/src/providers/provider.ts
```diff
@@ -140,13 +144,18 @@ export interface ThirdPartyProviderSupportsBuilds {
...
 export interface ThirdPartyProviderSupportsCodeAnalyzers {
 	fetchCodeAnalysis(
-		request: FetchThirdPartyCodeAnalyzersRequest
+		request: FetchThirdPartyCodeAnalyzersRequest,
+		params: IssueParams,
 	): Promise<FetchThirdPartyCodeAnalyzersResponse>;
+
+	fetchIsRepoMatch(
+		request: FetchThirdPartyRepoMatchToFossaRequest,
+	): Promise<FetchThirdPartyRepoMatchToFossaResponse>;
```
angieyeh, line 153: Change I made

shared/agent/src/providers/thirdPartyCodeAnalyzerProviderBase.ts
```diff
@@ -9,7 +9,7 @@ import {
 import { ThirdPartyProviderBase } from "./thirdPartyProviderBase";
 
 export abstract class ThirdPartyCodeAnalyzerProviderBase<
-		TProviderInfo extends CSProviderInfos = CSProviderInfos
+		TProviderInfo extends CSProviderInfos = CSProviderInfos,
```
angieyeh, line 12: prettier change

shared/ui/Authentication/actions.ts
```diff
@@ -54,7 +54,6 @@ import { localStore } from "../utilities/storage";
 import { emptyObject, uuid } from "../utils";
 import { HostApi } from "../webview-api";
 import { WebviewPanels } from "@codestream/webview/ipc/webview.protocol.common";
-import { setUserPreferences } from "../Stream/actions";
```
angieyeh, line 57: I think I borked my rebase, but this import is on line 51 so removing this line

shared/agent/src/providers/registry.ts
angieyeh, line 50: In fossa.ts there is one function called fetchCodeAnalysis, but in the `registry.ts` I split `fetchCodeAnalysis` into 2 functions, `fetchVulnerablities` and `fetchLicenseDependencies` . That is why `fetchVulnerablities` and `fetchLicenseDependencies` both use `FetchThirdPartyCodeAnalyzersRequest` and don't have their own request objects.

shared/agent/src/providers/registry.ts
```diff
@@ -819,14 +824,70 @@ export class ThirdPartyProviderRegistry {
...
+		const provider = getProvider(request.providerId);
+		if (provider === undefined) {
+			throw new Error(`No registered provider for '${request.providerId}'`);
+		}
+
+		const codeAnalyzersProvider = provider as ThirdPartyCodeAnalyzerProvider;
+		if (
+			codeAnalyzersProvider == null ||
+			typeof codeAnalyzersProvider.supportsCodeAnalysis !== "function" ||
+			!codeAnalyzersProvider.supportsCodeAnalysis()
```
angieyeh, line 878: Given this function is necessary for the CodeAnalysis code,  I was thinking this check is sufficient and I wouldn't need to create another supportsIsRepoMatch function. thoughts?

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -1,25 +1,37 @@
...
 export const CodeAnalyzers = (props: Props) => {
-	const [loading, setLoading] = useState(false);
-	const [message, setMessage] = useState("");
-	const [currentRepoId, setCurrentRepoId] = useMemoizedState<string | undefined>(undefined);
+	const [loading, setLoading] = useState<boolean | undefined>(undefined);
+	const [licDeploading, setLicDepLoading] = useState<boolean>(false);
+	const [vulnLoading, setVulnLoading] = useState<boolean>(false);
+	const [licenseDepIssues, setLicenseDepIssues] = useState<LicenseDependencyIssue[]>([]);
+	const [vulnIssues, setVulnIssues] = useState<VulnerabilityIssue[]>([]);
+	const [currentRepoId, setCurrentRepoId] = useMemoizedState<string | undefined | null>(null);
```
angieyeh, line 33: Left a codemark. I added a null state because repo's without remotes have undefined ids and the null state is helping determine when currentRepoId gets updated.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -32,62 +44,132 @@ export const CodeAnalyzers = (props: Props) => {
 				if (userProvider) providerInfo[name] = userProvider;
 			}
 		}
-		const currentRepoId = editorContext.scmInfo?.scm?.repoId;
-		const currentRepo = props.openRepos.find(_ => _.id === currentRepoId);
+		const activeFile = editorContext?.activeFile;
+		const currentRepo = props.openRepos.find(
+			_ => editorContext?.textEditorUri?.includes(_.folder.uri) && _.id === currentRepoId,
+		);
```
angieyeh, line 50: If currentRepoId is undefined it will match with the first repo with undefined repo id so I added an additional check that the current textEditorUri matches the openRepo folder uri. If there's a better way let me know!

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -32,62 +44,132 @@ export const CodeAnalyzers = (props: Props) => {
...
-	// The way providerInfo is created is causing useEffect
-	// to run in an infinite loop. Any new providers should be
-	// added to this memoized object.
-	const { fossa } = derivedState.providerInfo;
-	const memoizedProviderInfo = useMemo(
-		() => ({
-			fossa,
-		}),
-		[fossa],
-	);
```
angieyeh, line 55: Although providerInfo is memoized it is still causing useEffect to run twice every single time. This providerInfo object contains the user's access token information. So theoretically it shouldn't be triggering useEffect to run unless the token changes, but that is not happening.

I started to think why do we need this as a dependency. If a user changes the access token, that means they disconnected the integration and then entered their API token again. Can an access token change without this step? Does the server set a TTL for access tokens? 

Even if the server does have a TTL on access tokens, at the agent layer, `fossa.ts` gets the access token from `ThirdPartyProviderBase` so when an access token changes the UI will display whatever data it gets back. Maybe the user could get in a state where we display they have no Fossa issues when they do in fact have it, but I'm okay with that for now. If there are other reasons I should consider let me know!

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -32,62 +44,132 @@ export const CodeAnalyzers = (props: Props) => {
...
+					setLoading(false);
+				}
+			} catch (error) {
+				console.error("Error fetching data:", error);
+			}
+		};
+		fetchData();
+	}, [currentRepoId, derivedState.providers, props.paneState]);
+
+	const fetchMatchRepoToFossa = async (): Promise<boolean | undefined> => {
```
angieyeh, line 95: This function is new, but the diff makes it look like I modified it. I modified the function `fetchCodeAnalysis` which is now named `fetchLicenseDependencies`

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -98,16 +180,31 @@ export const CodeAnalyzers = (props: Props) => {
...
+						)}
+					{derivedState.bootstrapped &&
+						!loading &&
+						currentRepoId === undefined &&
+						isRepoMatched === false &&
+						!derivedState.currentRepo?.remotes?.length &&
+						derivedState.activeFile?.length && (
+							<div style={{ padding: "0 20px" }}>
+								Repo does not have a git remote, try another repo.
+							</div>
```
angieyeh, line 209: Still thinking about how to simplify these conditions for the conditional rendering. If you have any suggestions let me know!

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -0,0 +1,188 @@
...
+						return <LicenseDependencyRow licenseDependency={issue} />;
+					})}
+				</>
+			)}
+			{licenseDepExpanded && props.issues?.length === 0 && (
+				<Row style={{ padding: "0 10px 0 40px" }}>👍 No license dependency issues found</Row>
+			)}
+		</>
+	);
+});
```
angieyeh, line 188: There is a lot of copying and pasting in this file and I copied most of the code from `SecurityIssuesWrapper.tsx`. This was very quickly created so did not really think about reusability.

shared/util/src/protocol/agent/agent.protocol.fossa.ts
```diff
@@ -0,0 +1,102 @@
...
+		}
+	];
+	type: string;
+	vulnId?: string;
+	title?: string;
+	cve?: string;
+	cvss?: number;
+	severity?: string;
+	details?: string;
+	remediation?: string;
```
angieyeh, line 65: Because I made `fetchCodeAnalysis` in `fossa.ts` handle both LicenseDependencyIssues and VulnerabilityIssues  I had to make some of these attributes optional even though they most likely aren't optional. Is this a good way to do this?

shared/util/src/protocol/agent/agent.protocol.providers.ts
```diff
@@ -493,19 +494,70 @@ export const FetchThirdPartyBuildsRequestType = new RequestType<
...
 	void
 >("codestream/provider/codeAnalyzers");
 
+export interface FetchThirdPartyLicenseDependenciesRequest {
+	providerId: string;
+	repoId?: string;
+}
+export interface FetchThirdPartyLicenseDependenciesResponse {
+	issues?: LicenseDependencyIssue[];
+	error?: string; // CHECK THE TYPE! could be object
```
angieyeh, line 518: Error handling will be handled in a subsequent PR

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+const severityColorMap: Record<VulnSeverity, string> = {
+	CRITICAL: "#f52222",
+	HIGH: "#F5554B",
+	MEDIUM: "#F0B400",
+	LOW: "#0776e5",
+	UNKNOWN: "#ee8608",
+};
+
+function LicenseDependencyRow(props: { licenseDependency }) {
```
angieyeh, line 31: I was working on making things work so I removed all the TS types. I will add them in the next PR!

shared/util/src/protocol/agent/agent.protocol.providers.ts
```diff
@@ -493,19 +494,70 @@ export const FetchThirdPartyBuildsRequestType = new RequestType<
...
+}
+
+export const FetchThirdPartyLicenseDependenciesRequestType = new RequestType<
+	FetchThirdPartyLicenseDependenciesRequest,
+	FetchThirdPartyLicenseDependenciesResponse,
+	void,
+	void
+>("codestream/provider/licenseDependencies");
+
+export { LicenseDependencyIssue, VulnerabilityIssue } from "./agent.protocol.fossa";
```
angieyeh, line 528: Will remove this export in future PR. I didn't know how to export the interfaces out from this fossa specific file. Given we don't do this way of exporting anywhere in this file, I looked some more and it looks like we have another file called `agent.protocol.ts` where we export out provider specific interfaces. I will update the export to do it that way.. 

What are the differences between these files bc they both seem to export out similar information?

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
dsellarsnr: i think you can just use the CurrentRepoContext currentRepoCallback and maybe a useEffect when setCurrentRepoId changes.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
dsellarsnr: For the div with the same styling you can put the app logic into populating a message field and then here in the view logic you have just 1 check on `message &&...`

shared/util/src/protocol/agent/agent.protocol.providers.ts
dsellarsnr: One is vendor specific (agent.protocol.fossa.ts) - the other attempts to be a generic API that can work with any license / vulnerability vendor. Attempts is they key word - it is pretty difficult to do and usually gets messy.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: The useEffect doesn't use currentRepo only currentRepoId. I would like to remove currentRepo entirely and  I was so close, but the reason I kept it is because repos can have no remotes and in this case I want to show them a conditional message. The best way I could determine this state is using currentRepo.

shared/agent/src/providers/fossa.ts
```diff
@@ -47,15 +65,142 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
...
+		currentRepo: ReposScm,
+		fossaProjects: FossaProject[],
+		repoId?: string,
+	): Promise<FossaProject | Record<string, never>> {
+		if (repoId) {
+			for (const project of fossaProjects) {
+				let parsed;
+				try {
+					parsed = await GitRemoteParser.parseGitUrl(project.title);
+				} catch (ex) {}
```
eric-j, line 142: Might want to log an error here.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -32,60 +44,130 @@ export const CodeAnalyzers = (props: Props) => {
...
+			}
+		} catch (error) {
+			console.error(error);
+		}
+		return isRepoMatch;
+	};
+
+	const fetchVulnerabilities = async (): Promise<void> => {
+		if (vulnLoading) return;
+		setVulnLoading(true);
```
eric-j, line 120: Couldn't you return if !currentRepoId || vulnLoading, then setVunlLoading(true) afterwards?

Same for the logic in fetchLicenseDependencies below

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -98,16 +180,33 @@ export const CodeAnalyzers = (props: Props) => {
 			{props.paneState != PaneState.Collapsed && (
 				<PaneBody key="fossa">
 					{!derivedState.bootstrapped && <ConnectFossa />}
```
eric-j, line 182: Since every jsx conditional below this one must have bootstrapped === true, you could convert this to an if else to reduce some noise.

shared/agent/src/providers/fossa.ts
```diff
@@ -47,15 +65,142 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
...
 	get baseUrl() {
-		return `${this.apiUrl}/${this.apiPath}`;
+		return `${this.appUrl}${this.apiPath}`;
+	}
+
+	/**
+	 * Query string for Fossa Issues API
+	 * @returns query string
+	 */
+	private _getIssuesUrl(params: any) {
```
purplexa, line 75: It's not a big deal since it's a private method and the params are listed right below, but properly typing the parameter will give more useful information from the IDE when trying to use the method.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -32,60 +44,130 @@ export const CodeAnalyzers = (props: Props) => {
...
-	// to run in an infinite loop. Any new providers should be
-	// added to this memoized object.
-	const { fossa } = derivedState.providerInfo;
-	const memoizedProviderInfo = useMemo(
-		() => ({
-			fossa,
-		}),
-		[fossa],
-	);
+	useDidMount(() => {
```
purplexa, line 66: Is there a reason to use `useDidMount` here instead of just setting the default value of `loading` to true?

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -1,25 +1,37 @@
...
 	openRepos: ReposScm[];
 	paneState: PaneState;
 }
 
 export const CodeAnalyzers = (props: Props) => {
-	const [loading, setLoading] = useState(false);
-	const [message, setMessage] = useState("");
-	const [currentRepoId, setCurrentRepoId] = useMemoizedState<string | undefined>(undefined);
+	const [loading, setLoading] = useState<boolean | undefined>(undefined);
+	const [licDeploading, setLicDepLoading] = useState<boolean>(false);
```
purplexa, line 29: very minor, but did you mean to name this `licDepLoading` with a capital 'L'?

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -0,0 +1,188 @@
...
+				</div>
+				<Severity severity={criticalityToRiskSeverity(vuln.severity)} />
+			</Row>
+			{expanded && <VulnRow vuln={vuln} />}
+		</>
+	);
+}
+
+function VulnRow(props: { vuln }) {
+	const [expanded, setExpanded] = useState<boolean>(false);
```
purplexa, line 114: This isn't actually doing anything, is it?

shared/util/src/protocol/agent/agent.protocol.fossa.ts
purplexa: I'm not sure I understand the problem, but sometimes making a base type with the shared fields, then adding an additional field like `issueType` to each one with a fixed string can make the typing work. Something like:
```typescript
enum IssueType {
  A = "A",
  B = "B",
}

interface BaseType {
  sharedField: string;
}

interface TypeA extends BaseType {
  issueType: "A";
  fieldA: string;
}

interface TypeB extends BaseType {
  issueType: "B";
  fieldB: string;
}

type JoinedType = TypeA | TypeB;
```
Then you can do things like this without typescript complaining:
```typescript
function(param: JoinedType) {
  if (param.issueType === IssueType.A) return param.fieldA;
  if (param.issueType === IssueType.B) return param.fieldB;
  return undefined;
}
```
It's possible I messed something up but that's the general idea. Not sure if that's relevant to your problem, though.

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
angieyeh: 😅 it was copy pasta and doesn't do anything rn lol it will eventually be used for opening the modal. will remove since there's no modal yet.

shared/util/src/protocol/agent/agent.protocol.fossa.ts
angieyeh: that is exactly the problem and this approach is way cleaner, ty!

shared/util/src/protocol/agent/agent.protocol.providers.ts
angieyeh: would you say `agent.protocol.providers.ts` is the giant file where we put all the agent "routing"

shared/agent/src/providers/fossa.ts
angieyeh: that's a good reason to update!

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: yes, i was wondering about that, but then i left it bc CICD.tsx is written in a similar fashion. however since the code here is simpler would be good to just combine.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: previously there was some UI jank happening where we were testing if setting the initial loading state to be undefined would help, but i think the undefined is unnecessary now so i test what happens if we remove it

shared/agent/src/providers/fossa.ts
angieyeh: [6002f4b](https://github.com/TeamCodeStream/codestream/pull/1632/commits/6002f4b4d19ed410e579ff392c030c577597cbcb)

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: [b089669](https://github.com/TeamCodeStream/codestream/pull/1632/commits/b08966999fd002db73f9f4677442dc8287b2c27c)
there are other UI changes in commit, but removed it

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: [b089669](https://github.com/TeamCodeStream/codestream/pull/1632/commits/b08966999fd002db73f9f4677442dc8287b2c27c)
there are other UI changes in commit, but updated it

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
angieyeh: [b089669](https://github.com/TeamCodeStream/codestream/pull/1632/commits/b08966999fd002db73f9f4677442dc8287b2c27c)
there are other UI changes in commit, but removed it

shared/util/src/protocol/agent/agent.protocol.fossa.ts
angieyeh: [6ee34f6](https://github.com/TeamCodeStream/codestream/pull/1632/commits/6ee34f60e8b7c1f2384548431966f0faeee4f160)
making the attributes in the interfaces not optional, required splitting out fetchCodeAnalysis in fossa.ts into two functions fetchLicenseDepencies and fetchVulnerabilities. if i didn't split it out, all the code that called fetchCodeAnalysis would require setting the result to have the type VulnerabilityIssues[] | LicenseDependencyIssues[] regardless if the request would only fetch one of the given types. the reason i originally wanted one function was so i wouldn't have two functions that had almost identical code, but i think having almost duplicate code is better than having to always define type with the OR condition.

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: [b089669](https://github.com/TeamCodeStream/codestream/pull/1632/commits/b08966999fd002db73f9f4677442dc8287b2c27c)
there are other UI changes in commit, but created a function for the conditional text

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
angieyeh: [b089669](https://github.com/TeamCodeStream/codestream/pull/1632/commits/b08966999fd002db73f9f4677442dc8287b2c27c)
i didn't use a ternary, but worked on cleaning up the code

shared/agent/src/providers/fossa.ts
angieyeh: [6ee34f6](https://github.com/TeamCodeStream/codestream/pull/1632/commits/6ee34f60e8b7c1f2384548431966f0faeee4f160)
the change got committed as part of a larger change

shared/ui/Stream/CodeAnalyzers/CodeAnalyzers.tsx
```diff
@@ -170,6 +172,23 @@ export const CodeAnalyzers = (props: Props) => {
...
+		if (currentRepoId && isRepoMatched === false) {
+			return "Project not found on FOSSA";
+		}
+		if (derivedState.hasRemotes === false && derivedState.hasActiveFile) {
+			return "Repo does not have a git remote, try another repo.";
+		}
+		return "";
+	};
+
+	const hasConditonalText = conditionalText();
```
purplexa, line 190: minor typo if you care: hasCondit***i***onalText

purplexa (approved):

shared/agent/src/providers/fossa.ts
```diff
@@ -47,15 +65,142 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
...
+
+	/**
+	 * Matches current repo to Fossa project
+	 * @returns project object
+	 */
+	private async _matchRepoToFossaProject(
+		currentRepo: ReposScm,
+		fossaProjects: FossaProject[],
+		repoId?: string,
+	): Promise<FossaProject | Record<string, never>> {
```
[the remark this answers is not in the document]
angieyeh: how about `object` instead of `Record<string, never>`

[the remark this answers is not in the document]
purplexa: I think Dillon was suggesting you change the return value on the unsuccessful case from `{}` to `undefined`, which shouldn't make using it any more complicated.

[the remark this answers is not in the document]
angieyeh: 01fa093

PR #1638, opened 2023-07-26
Anomaly detection notifications and support for more languages

marcelofarias:


vs/src/CodeStream.VisualStudio.Shared/Services/SymbolService.cs
```diff
@@ -0,0 +1,74 @@
...
+					if (project.SupportsCompilation)
+					{
+						var symbols = (await SymbolFinder
+							.FindDeclarationsAsync(project, methodName, true))
+							.Where(x => x.ToDisplayString().EqualsIgnoreCase($"{fullyQualifiedMethodName}()"));
+						
+						if (symbols != null && symbols.Any())
+						{
+							var symbol = symbols.FirstOrDefault();
+							await _workspace.TryGoToDefinitionAsync(symbol, project, cancellationToken);
```
bcanzanella, line 65: @CalvinAllen does this need to be awaited? can it fire and forget? i forget the syntax... `_ = _workspace.Try...` ?

shared/agent/test/unit/providers/newrelic/newrelic.clm.queries.spec.ts
```diff
@@ -9,17 +9,17 @@ describe("clm query generation", () => {
 				"Nested/Controller/agents/update",
 			]);
 			expect(query).toContain(
-				"metricTimesliceName in ('Controller/agents/show','Controller/agents/update')"
+				"metricTimesliceName in ('Controller/agents/show','Controller/agents/update')",
 			);
 		});
 
-		it("adds WebTransaction/ path for python", () => {
+		xit("adds WebTransaction/ path for python", () => {
```
bcanzanella, line 16: ??

marcelofarias: Based on all the data I saw, we should get the numbers from Function/* rather than WebTransaction/Function/*. Also: https://chat.openai.com/share/04814a64-5872-4665-a427-2b672e2a7297

bcanzanella: i think my question was more of "what is `xit`?"

bcanzanella: got it, disabled.

vs/src/CodeStream.VisualStudio.Shared/Services/SymbolService.cs
CalvinAllen: @bcanzanella Not certain; I can play around with it and see - but yeah, I believe that's one variation of doing it, IIRC

bcanzanella: def a nice-to-have, was just thinking what we can do to eek out as much perf from VS as we can.

dsellarsnr (approved):
🌟

vs/src/CodeStream.VisualStudio.Shared/Services/SymbolService.cs
CalvinAllen: Done.

PR #1646, opened 2023-07-27
[FOSSA] Add Vulnerability and License Dependency Details View, and Error Handling

angieyeh:
This PR handles NR-119857 which is for adding modal details view. Also did clean up for imports and exports, and added error handling.

Vuln modal
Screenshot 2023-07-26 at 5 13 02 PM
Lic Dep modal
Screenshot 2023-07-26 at 5 13 15 PM
Fetch vuln and lic dep error
Screenshot 2023-07-26 at 5 11 20 PM
Only vuln error
Screenshot 2023-07-27 at 11 28 02 AM
Only lic dep error
Screenshot 2023-07-27 at 11 29 06 AM
No repo match error
Screenshot 2023-07-27 at 12 13 28 PM

### DIFF
--- a/shared/agent/src/providers/fossa.ts
+++ b/shared/agent/src/providers/fossa.ts
@@ -177,7 +177,10 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 			return { isRepoMatch: true };
 		} catch (error) {
 			Logger.error(error);
-			return { error: error.message };
+			if (error.message === "Unauthorized") {
+				return { error: "API token is invalid" };
+			}
+			return { error: "Project not found on FOSSA" };
 		}
 	}
 
@@ -208,9 +211,7 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
 			};
 		} catch (error) {
 			Logger.error(error);
-			return {
-				error,
-			};
+			return { error: "Error fetching issues from FOSSA" };
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


angieyeh:
Leaving it as draft pr until https://github.com/TeamCodeStream/codestream/pull/1632 gets merged.

shared/agent/src/providers/fossa.ts
```diff
@@ -167,16 +164,24 @@ export class FossaProvider extends ThirdPartyCodeAnalyzerProviderBase<CSFossaPro
...
+			}
+			const projects: FossaProject[] = await this._getProjects();
+			const project = await this._matchRepoToFossaProject(currentRepo, projects, request.repoId);
+			if (!project) {
+				return { isRepoMatch: false };
+			}
+			return { isRepoMatch: true };
+		} catch (error) {
+			Logger.error(error);
+			if (error.message === "Unauthorized") {
```
eric-j, line 180: Would be safer to use an error code here if available.  


If none is available from fossa error response, then maybe standardize the string to lower case before doing equals check.

eric-j (approved):
one small thing otherwise :+1:

shared/agent/src/providers/fossa.ts
angieyeh: when i make a request in postman there is a property called httpStatusCode, but when i tried to access that property on the error object it didn't exist 🤔 hence the equality check. standardizing sounds good and would make it safer thanks!

angieyeh: after reading some more about JS errors, when an error is caught, the error variable that gets passed to the catch is the built in JS error object which contains the name, message and stack properties. it seems like behind the scenes, JS is parsing the error response we get back from fossa. is there anyway to access the httpStatusCode, or is the built in JS error object the best we can get?

PR #1647, opened 2023-07-28
[FOSSA] Add event tracking for FOSSA

angieyeh:
PR handles NR-119857. We are adding event tracking for FOSSA vulnerability and license dependency clicks.

### DIFF
--- a/shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
+++ b/shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
@@ -5,6 +5,7 @@ import {
 	VulnSeverity,
 	VulnerabilityIssue,
 } from "@codestream/protocols/agent";
+import { HostApi } from "@codestream/webview/webview-api";
 import Icon from "../Icon";
 import Tooltip from "../Tooltip";
 import { Link } from "@codestream/webview/Stream/Link";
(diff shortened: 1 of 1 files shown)


shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -97,9 +98,11 @@ function ModalView(props: {
 	displays: (string | boolean)[][];
 	title: string;
 	details: string;
+	trackingData: { "Analyzer Service": string; Category: string };
 	onClose: () => void;
 }) {
-	const { displays, title, details } = props;
+	const { displays, title, details, trackingData } = props;
+	HostApi.instance.track("Analyzer Result Clicked", trackingData);
```
eric-j, line 105: Since the tracking subject is "Analyzer Result Clicked", it might make sense to have this api call in the onClick function where we set expanded to true.


On that note, is expanded in this case expanding a row?  Or is it opening a modal?  If its the latter, you might want to come up with a more accurate description for that variable.

angieyeh: Those are good points. I didn't want to deviate from what we currently do so I followed what is being done in VM in SecurityIssuesWrapper.tsx. When the modal is clicked we fire off the event with the message "Vulnerability Clicked".

I think the message is fine based off of what we do in VM. However I will check with Dave to see if he prefers more specificity and perhaps more events.

angieyeh: I talked to Dave about the event tracking and the message itself is fine how it is. In regards to where we should call the event tracking,  in the modal or the on click handler, the event is to track that the modal has been viewed and not that row was clicked to open the modal. The message is kind of a misnomer, but given what we're tracking, I think leaving the API call where it is is fine. Thoughts?

eric-j (approved):

angieyeh:
I will have to merge https://github.com/TeamCodeStream/codestream/pull/1646 into feature/fossa and it should resolve most if not all of these conflicts.

PR #1651, opened 2023-08-01
Allow admins to delete posts

reneelpetit:
DONE


shared/ui/Stream/Codemark.tsx
```diff
@@ -1360,7 +1363,7 @@ export class Codemark extends React.Component<Props, State> {
 			});
 		}
 
-		if (mine) {
+		if (mine || isAdmin) {
```
purplexa, line 1366: I don't think admins are allowed to edit other people's codemarks on the backend, so you probably want to modify this to only show the delete option for admins, not the edit option.

reneelpetit: Good point!  You're right, the ticket only mentions delete ability not edit. I'll fix that

purplexa (approved):

PR #1662, opened 2023-08-03
fix allow admin to delete connect MapStateToProps

reneelpetit:
https://issues.newrelic.com/browse/NR-101485

### DIFF
--- a/shared/ui/Stream/Codemark.tsx
+++ b/shared/ui/Stream/Codemark.tsx
@@ -45,6 +45,7 @@ import { getPosts } from "../store/posts/actions";
 import { getPost } from "../store/posts/reducer";
 import { getReview } from "../store/reviews/reducer";
 import {
+	currentUserIsAdminSelector,
 	getTeamMembers,
 	getTeamTagsHash,
 	getUserByCsId,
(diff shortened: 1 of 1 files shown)


shared/ui/Stream/Codemark.tsx
```diff
@@ -1360,11 +1362,15 @@ export class Codemark extends React.Component<Props, State> {
 			});
 		}
 
-		if (mine) {
-			menuItems.push(
-				{ label: "Edit", action: () => this.setState({ isEditing: true }) },
-				{ label: "Delete", action: this.deleteCodemark }
-			);
+		if (mine || connect(this.props.currentUserIsAdminSelector)) {
```
purplexa, line 1365: Does this actually work? `connect()` should return a function, so I think this will just always be true.

`connect()` primarily takes two arguments: `mapStateToProps` and `mapDispatchToProps`. It's used to inject state values and dispatched methods as props so they're available in a component by returning a helper function you can call, passing in the component class. I believe what you want to do here is inside the `mapStateToProps` method near the bottom of the file, add in a line where you set a value, something like `isCurrentUserAdmin`, to the result of `currentUserIsAdminSelecter(state)`. Then return that value in the return object, and add an appropriate value to the `ConnectedProps` type above. That way you should be able to work with the value directly here.

Reference: https://react-redux.js.org/api/connect

reneelpetit: OK, I think I got it now!  I had already looked at that article you linked before but then I watched a video and finally I understood.  I think it's set up correctly now.

purplexa (approved):

dsellarsnr (commented):
looks good

PR #1665, opened 2023-08-04
Pre-load vulnerabilities: NR-108350

reneelpetit:


shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -322,6 +323,15 @@ export const SecurityIssuesWrapper = React.memo((props: Props) => {
 		[error]
 	);
 
+	if (data && data.totalRecords > 0) {
+		props.setVal(true);
+	}
+
+	const warningTooltip =
+		data && data.totalRecords > 0 && data.totalRecords === 1
```
bcanzanella, line 331: can be simplified to `data && data.totalRecords === 1`

reneelpetit: oh yeah good point!

shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -30,6 +30,7 @@ interface Props {
 	currentRepoId: string;
 	entityGuid: string;
 	accountId: number;
+	setVal: Function;
```
bcanzanella, line 33: nitpick on the name, but what does "setVal" mean here? seems perhaps too generic?

reneelpetit: yes good point, I'll make it less generic and more meaningful

shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -30,6 +30,7 @@ interface Props {
 	currentRepoId: string;
 	entityGuid: string;
 	accountId: number;
+	setHasVulnerabilities: Function;
```
purplexa, line 33: probably worth being more specific and setting the type to `(value: boolean) => void`

shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -322,6 +323,13 @@ export const SecurityIssuesWrapper = React.memo((props: Props) => {
 		[error]
 	);
 
+	if (data && data.totalRecords > 0) {
+		props.setHasVulnerabilities(true);
```
purplexa, line 327: If I'm reading the docs right, I don't think you're supposed to call state setters directly during the render, at least not in a component other than the one where the state was defined. I think you're supposed to wrap this in a useEffect, like

```typescript
useEffect(() => {
	if (data && data.totalRecords > 0) {
		props.setHasVulnerabilities(true);
	}
}, [data, props.setHasVulnerabilities]);
```

purplexa (approved):

PR #1673, opened 2023-08-10
[FOSSA] Remove Extra Level in Vulnerability Results Tree

angieyeh:
PR handles NR-149766

It's not worth it to have each vuln expandable so that the user can see the title. Users can click the vuln and see the title in the details view. Also added the vulnerability title as the tooltip instead of the recommended fix.

Before:
Screenshot 2023-08-10 at 2 34 55 PM

After:

https://github.com/TeamCodeStream/codestream/assets/14142356/c97b3acf-133d-4c44-8838-1ce08d0e7809


eric-j (approved):
👍

dsellarsnr (approved):

PR #1678, opened 2023-08-11
[FOSSA] Add Event Tracking Try 2

angieyeh:


angieyeh:
No idea what happened in https://github.com/TeamCodeStream/codestream/pull/1647 but that change seemed to not make it into feature/fossa

PR #1681, opened 2023-08-14
Feature/grok streaming sequence

dsellarsnr:
Use sequence numbers to correclty order the pubnub events for the grok stream

### DIFF
--- a/shared/ui/store/codeErrors/thunks.tsx
+++ b/shared/ui/store/codeErrors/thunks.tsx
@@ -52,6 +52,7 @@ import React from "react";
 import { getGrokPostLength } from "@codestream/webview/store/posts/reducer";
 import { URI } from "vscode-uri";
 import { clearResolvedFlag } from "@codestream/utils/api/codeErrorCleanup";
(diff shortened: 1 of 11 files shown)


shared/ui/store/codeErrors/thunks.tsx
```diff
@@ -671,20 +672,15 @@ export const handleGrokError = (grokError: CSAsyncGrokError) => async dispatch =
...
-	if (doneEvent) {
-		dispatch(setGrokLoading(false));
-	}
+export const handleGrokChonk = (events: CSGrokStream[]) => async dispatch => {
+	if (events.length === 0) return;
+	const grokStoreEvents: GrokStreamEvent[] = events.map(e => ({
+		sequence: e.sequence,
+		postId: e.extra.postId,
+		streamId: e.extra.streamId,
+		content: e?.content?.content,
```
bcanzanella, line 681: should other `e`'s be guarded?

dsellarsnr: They are all required fields, always sent on server. Only content is optional for the "done" event case.

bcanzanella (approved):

PR #1696, opened 2023-08-18
track event fires once after click

reneelpetit:
https://issues.newrelic.com/browse/NR-121157

### DIFF
--- a/shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
+++ b/shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
@@ -73,6 +73,12 @@ const criticalityToRiskSeverity = (riskSeverity: VulnSeverity): VulnSeverity => 
 	}
 };
 
+const eventClicked = (
+	trackingData: { "Analyzer Service": string; Category: string } | undefined
+) => {
+	HostApi.instance.track("Analyzer Result Clicked", trackingData);
+};
+
 const ModalView = (props: {
 	displays: {
 		label: string;
@@ -81,11 +87,10 @@ const ModalView = (props: {
 	}[];
 	title: string;
 	details: string;
-	trackingData: { "Analyzer Service": string; Category: string };
+
 	onClose: () => void;
 }) => {
-	const { displays, title, details, trackingData } = props;
-	HostApi.instance.track("Analyzer Result Clicked", trackingData);
+	const { displays, title, details } = props;
 
 	return (
 		<div className="codemark-form-container">
@@ -220,6 +225,7 @@ const VulnerabilityRow = (props: { issue: VulnerabilityIssue }) => {
 				className={"pr-row"}
 				onClick={() => {
 					setExpanded(!expanded);
+					eventClicked({ "Analyzer Service": "FOSSA", Category: "Vulnerability" });
 				}}
 			>
 				<Tooltip placement="bottom" title={tooltipText} delay={1}>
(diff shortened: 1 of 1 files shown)


shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -220,6 +225,7 @@ const VulnerabilityRow = (props: { issue: VulnerabilityIssue }) => {
 				className={"pr-row"}
 				onClick={() => {
 					setExpanded(!expanded);
+					eventClicked({ "Analyzer Service": "FOSSA", Category: "Vulnerability" });
```
purplexa, line 228: I'm thinking it probably makes sense to only fire the event if `expanded` is `false`, otherwise if you click it to expand then click it again to hide the details that would result in two events firing when it was only one vulnerability clicked, which might make the data slightly less accurate.

reneelpetit: Yeah that would make sense, but the it doesn't happen that way.  The way it's currently set up, when you open the details it's that onClick that gets triggered, and when you close the details, it's a different, separate part of the code that gets triggered, the onClose. It doesn't actually trigger twice, only once when opening.  I could clean up the code that's there, though, so it's more of a toggle instead of separate processes?

reneelpetit: I didn't make it a toggle but I made it more clear what's happening by just using true & false

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -219,7 +222,10 @@ const VulnerabilityRow = (props: { issue: VulnerabilityIssue }) => {
 				style={{ padding: "0 10px 0 30px" }}
 				className={"pr-row"}
 				onClick={() => {
-					setExpanded(!expanded);
+					if (!expanded) {
+						eventClicked({ "Analyzer Service": "FOSSA", Category: "Vulnerability" });
+					}
+					setExpanded(true);
```
bcanzanella, line 228: would still keep this `setExpanded(!expanded)` since if this later changes, the toggle-ness of it continues to work.

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -259,7 +264,6 @@ const VulnerabilityRow = (props: { issue: VulnerabilityIssue }) => {
 							},
 							{ label: "References", description: vuln.references, link: true },
 						]}
-						onClose={() => setExpanded(false)}
```
bcanzanella, line 262: does this need to return?

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
```diff
@@ -314,7 +318,6 @@ const LicenseDependencyRow = (props: { issue: LicenseDependencyIssue }) => {
 								description: licenseDependency.depths.direct ? "Direct" : "Transitive",
 							},
 						]}
-						onClose={() => setExpanded(false)}
```
bcanzanella, line 317: does this need to return?

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
reneelpetit: looks like it's set up the way it is because of the props requirement for the <Modal>

shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
reneelpetit: looks like it's set up the way it is because of the props requirement for the <Modal>

bcanzanella (approved):

PR #1712, opened 2023-08-31
Remove “Auto hide markers”, “Show avatars”, “Show Marker Code Lens”, and “Show Shortcut Tip On Selection” settings.

CalvinAllen:
Default "Show Avatars" to true (for now), and since it was easy enough to do, all the code revolving "Marker Code Lens" and "Show Shortcut Tip" was removed entirely (other than defaulting to False)


dsellarsnr (approved):
✂️

PR #1730, opened 2023-09-05
distributed tracing and min agent version checks for anom detection

eric-j:


shared/util/src/protocol/agent/api.protocol.models.ts
```diff
@@ -901,6 +901,16 @@ export const DEFAULT_CLM_SETTINGS: CLMSettings = {
 	minimumAverageDurationValue: "0.1",
 };
 
+export const REQUIRED_AGENT_VERSIONS = {
+	go: "3.24.0",
+	java: "7.11.0",
+	".net": "10.2.0",
+	"node.js": "10.5.0",
+	php: "10.6.0 ",
+	python: "7.10.0.175",
```
dsellarsnr, line 910: This python version is not a valid semver - will throw `TypeError: Invalid Version: 7.10.0.175`

Can probably workaround by using semver.coerce on the actual agent version and this version (it will chop off the ending digits)

shared/util/src/protocol/agent/api.protocol.models.ts
```diff
@@ -901,6 +901,16 @@ export const DEFAULT_CLM_SETTINGS: CLMSettings = {
 	minimumAverageDurationValue: "0.1",
 };
 
+export const REQUIRED_AGENT_VERSIONS = {
```
dsellarsnr, line 904: This probably doesn't belong in api.protocol - can just live in newrelic.ts

PR #1760, opened 2023-09-15
Unexpected Error when open an error via Activity Feed or permalink NR-146780

reneelpetit:
Works now, but I need to fix all the //@ts-ignore

### DIFF
--- a/shared/agent/src/providers/newrelic.ts
+++ b/shared/agent/src/providers/newrelic.ts
@@ -1672,111 +1672,126 @@ export class NewRelicProvider
 				}
 			);
 
-			if (errorGroupFullResponse?.actor?.errorsInbox?.errorGroups?.results?.length) {
-				const errorGroupResponse = errorGroupFullResponse.actor.errorsInbox.errorGroups.results[0];
-				entityGuid = errorGroupResponse.entityGuid;
-				errorGroup = {
-					entity: {},
-					accountId: accountId,
-					entityGuid: entityGuid,
-					guid: errorGroupResponse.id,
-					title: errorGroupResponse.name,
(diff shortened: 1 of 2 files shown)


shared/agent/src/providers/newrelic.ts
```diff
@@ -3102,7 +3117,62 @@ export class NewRelicProvider
 					ids: [errorGroupGuid],
 				}
 			);
-			return response?.actor?.errorsInbox?.errorGroups?.results[0] || undefined;
+			if (response?.actor?.errorsInbox?.errorGroups?.results[0]) {
+				return response?.actor?.errorsInbox?.errorGroups?.results[0];
+			}
+			//try with no timestamp info
+			try {
```
bcanzanella, line 3124: @reneelpetit if the timestamp arg is missing, should it call this directly instead of calling the first, it not finding anything, then calling this? (aka making 2 queries instead of 1)

shared/agent/src/providers/newrelic.ts
```diff
@@ -3289,11 +3360,92 @@ export class NewRelicProvider
...
+				accountId: accountId,
+				errorGroupGuids: [errorGroupGuid],
+				entityGuid: entityGuid,
+			});
+		} catch (ex) {
+			Logger.error(ex);
+		}
+
+		try {
+			response = this.query(q2, {
```
bcanzanella, line 3439: similar question here

shared/agent/src/providers/newrelic.ts
bcanzanella: also, did you get to the root of the issue of why certain times there is no timestamp?

PR #1762, opened 2023-09-18
fossa remediation advice logic NR-162079

reneelpetit:
This is the block coming in:
```
"remediation": {
                "partialFix": "5.7.2",
                "partialFixDistance": "PATCH", (this key wont be present if they cant determine if PATCH, MINOR, MAJOR, etc)
                "completeFix": "5.7.2",
                "completeFixDistance": "PATCH" (also might not be set)
            }
```
The logic we want to do here is:
```
// if these are the same, the partial and complete are the same
if (remediation.partialFix.equals(remediation.completeFix)) {
     if (isset(remediation.partialFixDistance)) {
                display remediation.partialFixDistance + ":" + remediation.partialFix
      } else {
                 display remediation.partialFix
      }
} else {
// we have both a partial and a complete fix, so we need to display both
      display 'Partial Remediation:' + isset(remediation.partialFixDistance) ? remediation.partialFixDistance : "" + remediation.partialFix
      display 'Complete Remediation:' + isset(remediation.completeFixDistance) ? remediation.completeFixDistance : "" + remediation.completeFix
}
```

### DIFF
--- a/shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
+++ b/shared/ui/Stream/CodeAnalyzers/FossaIssues.tsx
@@ -211,10 +211,26 @@ const VulnerabilityRow = (props: { issue: VulnerabilityIssue }) => {
 	const [expanded, setExpanded] = useState<boolean>(false);
 	const vuln = props.issue;
 
+	function remediationFix() {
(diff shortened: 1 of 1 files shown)


dsellarsnr (approved):
👍

PR #1763, opened 2023-09-19
Move 'shared/ui' to '/shared/webviews/sidebar'

CalvinAllen:
This isn't 100% ready, but had to move and rename A LOT of files - feel free to peruse early and often :)


CalvinAllen:
This has been caught up to develop as of today. A lot of the file "changes" are formatting - I guess because I moved them around so they got picked up by the pre-commit hook?

PR #1800, opened 2023-10-12
Renames from "CodeStream" to "New Relic CodeStream"

CalvinAllen:


vs/src/CodeStream.VisualStudio.Core/Application.cs
```diff
@@ -13,7 +13,7 @@ namespace CodeStream.VisualStudio.Core
 {
 	public class Application
 	{
-		public const string Name = "CodeStream";
+		public const string Name = "New Relic CodeStream";
```
bcanzanella, line 16: Will this affect the Credentials Service?

CalvinAllen: Good call, I'm not sure. I definitely had to login, but I attributed it to changing the settings path (which I reverted)

CalvinAllen: Crap, yes it did.

bcanzanella (approved):

PR #1801, opened 2023-10-13
removes documentChanged notifications

bcanzanella:
I've been meaning to get this in. This removes some "chatty" notifications around documents opening/changing/saving/closing that was used exclusively for some features that are no longer used (namely FRs and the people/change screen).


eric-j (approved):
👍 nice

PR #1804, opened 2023-10-13
updated ruby extension dependency

eric-j:
image


dsellarsnr:
Don't forget src/providers/instrumentationCodeLensProvider.ts and src/__test__/suite/InstrumentationCodeLensProvider.test.ts

For me it required rvm to be installed but it is an ugly install process so maybe we should just link to rvm home page at https://rvm.io/rvm/install

dsellarsnr (approved):

PR #1807, opened 2023-10-16
NR-161545 Upgrade node.js to 18.15.0

dsellarsnr:


bcanzanella:
@dsellarsnr check in Node.cs public static bool EnsureVersion() has a version check on node.

also, when this gets in, please ask all the devs to upgrade their node versions

bcanzanella:
also, will TC need any changes for this?

dsellarsnr:
Ah I didn't realize how the vs node worked. Looks like i can upgrade it from 16 to 18 in agent/package.json via pkg

dsellarsnr:
TeamCity changes already done

PR #1814, opened 2023-10-17
Feature/auto token refresh

cstryker:
This is intended for the feature/uid branch (so can't be merged to develop). The problem this is trying to solve is that when the New Relic access token is issued, it needs a refresh within an hour. For other providers, a refresh-on-demand approach is taken, but this doesn't work for the NR access token, because if an hour goes by, and the token is already expired, and the user refreshes their IDE without doing anything else O11y related, the token won't pass through Service Gateway and the auto-login on refresh will fail. So I've implemented a timer-based solution to refresh the token automatically when it is close to expiration. Seeing as this might be useful for other providers, I've done this in the generic case, not just for NR.


shared/agent/src/providers/thirdPartyProviderBase.ts
```diff
@@ -278,14 +285,22 @@ export abstract class ThirdPartyProviderBase<
 				return;
 			}
 
-			const oneMinuteBeforeExpiration = this._providerInfo.expiresAt - 1000 * 60;
-			if (oneMinuteBeforeExpiration > new Date().getTime()) return;
+			const expirationTriggerTime =
+				this._providerInfo.expiresAt - 1000 * TOKEN_EXPIRATION_TOLERANCE_SECONDS;
```
bcanzanella, line 289: does this._providerInfo.expiresAt get updated via some other means? pubnub?

cstryker: Yes, when the token is refreshed (server-side), a PubNub message updates the providerInfo on the client side.

dsellarsnr:
A question and a concern.

What happens when laptop off network / turned off overnight? Will user have to sign in from scratch next morning? 

Most of our timers are not too time sensitive as they just do polling - we can tolerate what happens when the OS sleeps without worrying about when exactly the timer fires. My best understanding from this [stack overflow](https://stackoverflow.com/questions/6346849/what-happens-to-settimeout-when-the-computer-goes-to-sleep) is that the node timers pretend the sleep never happened - so if you have 30 minutes left on your timer, laptop goes to sleep for 45 minutes, when the laptop wakes up it will still have 30 minute until the timer callback happens for a total time of 1 hour 15 minutes. This might be a case where we have to poll every minute and check against the current wall clock time.

cstryker:
> A question and a concern.
> 
> What happens when laptop off network / turned off overnight? Will user have to sign in from scratch next morning?
> 
> Most of our timers are not too time sensitive as they just do polling - we can tolerate what happens when the OS sleeps without worrying about when exactly the timer fires. My best understanding from this [stack overflow](https://stackoverflow.com/questions/6346849/what-happens-to-settimeout-when-the-computer-goes-to-sleep) is that the node timers pretend the laptop sleep never happened - so if you have 30 minutes left on your timer, laptop goes to sleep for 45 minutes, when the laptop wakes up it will still have 30 minute until the timer callback happens for a total time of 1 hour 15 minutes. This might be a case where we have to poll every minute and check against the current wall clock time.

@dsellarsnr Yeah, unfortunately that does appear to be the case. So yeah, I think you're once-a-minute solution is the way to go. I'll work on that and resubmit.

PR #1827, opened 2023-10-30
Feature/newrelic token refresh on demand

cstryker:
An expired token is checked, and an attempt is made to refresh, in three places: (1) on IDE refresh (resulting in a "token login"), (2) on a CodeStream API request, (3) on a NerdGraph request. In all cases, the token needs to be updated in (1) the "token manager" in VSCode (for IDE refresh), (2) the codestreamApi object (for CodeStream API requests), and (3) the providerInfo structure for newrelic provider (for NerdGraph requests). This scattering of the token kinda sucks, but would require a much more radical code rewrite to fix.

The VSCode related changes here will need to propagated to JB and VS.


shared/agent/src/api/codestream/codestreamApi.ts
```diff
@@ -2411,47 +2439,40 @@ export class CodeStreamApiProvider implements ApiProvider {
...
 
-		try {
-			// For refresh of token behind Service Gateway, use a separate no-auth request to pass through
-			// Service Gateway without authentication
-			let response;
-			if (providerId === "newrelic" && this._usingServiceGatewayAuth) {
+		if (this._refreshNRTokenPromise) {
+			return this._refreshNRTokenPromise;
+		}
+		this._refreshNRTokenPromise = new Promise(async resolve => {
```
dsellarsnr, line 2449: Long story short, async Promise executor is [not great](https://stackoverflow.com/questions/43083696/cant-throw-error-from-within-an-async-promise-executor-function). Here is a version using the promise style

```
	@log()
	refreshNewRelicToken(refreshToken: string): Promise<CSNewRelicProviderInfo> {
		const cc = Logger.getCorrelationContext();

		if (this._refreshNRTokenPromise) {
			return this._refreshNRTokenPromise;
		}
		this._refreshNRTokenPromise = new Promise((resolve, reject) => {
			const url = "/no-auth/provider-refresh/newrelic";
			this.put<{ refreshToken: string }, CSNewRelicProviderInfo>(url, {
				refreshToken,
			})
				.then(response => {
					if (response.accessToken) {
						Logger.log("New Relic access token successfully refreshed, setting...");
						this.setAccessToken(response.accessToken, {
							expiresAt: response.expiresAt!,
							refreshToken: response.refreshToken!,
						});
						if (SessionContainer.isInitialized()) {
							SessionContainer.instance().session.onAccessTokenChanged(
								response.accessToken,
								response.refreshToken
							);
						}
					}
					delete this._refreshNRTokenPromise;
					resolve(response);
				})
				.catch(ex => {
					Logger.error(ex, cc);
					reject(ex);
				});
		});
		return this._refreshNRTokenPromise;
	}
```

shared/agent/src/providers/newrelic.ts
```diff
@@ -487,6 +487,40 @@ export class NewRelicProvider
 		};
 	}
 
+	private async clientRequestWrap<T>(query: string, variables: any, useOtherRegion?: boolean) {
+		let resp;
+		const client = await this.client(useOtherRegion);
+		let triedRefresh = false;
+		while (!resp) {
+			try {
+				resp = await (await this.client()).request<T>(query, variables);
```
dsellarsnr, line 496: Should this use the client returned on line 492?

CalvinAllen:
I added the VS implementation to this branch if anybody wants to review (@cstryker / @dsellarsnr) since Brian is out. Don't have an "update token" method, so delete and add

shared/agent/src/api/codestream/codestreamApi.ts
cstryker: Good call. Will fix.

shared/agent/src/providers/newrelic.ts
cstryker: Oops. Yes. Good catch.

cstryker:
@dsellarsnr made the requested changed. Took a look at Calvin's VS stuff and seems OK to me but could probably use a second pair of eyes.

shared/agent/src/session.ts
```diff
@@ -1312,6 +1322,11 @@ export class CodeStreamSession {
 			this.listenForEchoes();
 		}
 
+		if (this.api.usingServiceGatewayAuth) {
+			Logger.log("Service Gateway authentication is active, changing access token...");
+			this.onAccessTokenChanged(token.value, response.accessTokenInfo?.refreshToken);
+		}
+
```
dsellarsnr, line 1329: When debugging I found the token was saved in the IDE twice in a row - I think like 1313 with `this.agent.sendNotification(DidLoginNotificationType, { data: loginResponse });` is enough and you don't need to fire onAccessTokenChanged.

cstryker: Fair enough. Fixed.

dsellarsnr (approved):

PR #1830, opened 2023-11-03
Fix slow operations are prohibited on EDT in JB 2023.3 EAP

dsellarsnr:


jb/src/main/kotlin/clm/CLMEditorManager.kt
```diff
@@ -51,13 +52,15 @@ import com.intellij.psi.SyntaxTraverser
 import com.intellij.refactoring.suggested.endOffset
 import com.intellij.refactoring.suggested.startOffset
 import com.intellij.util.concurrency.NonUrgentExecutor
+import com.jetbrains.rd.framework.util.withContext
```
marcelofarias, line 55: This might throw a runtime error

dsellarsnr: ugh it keeps happening

PR #1833, opened 2023-11-09
Adds NR events

bcanzanella:
This adds NR event tracking (along with the current provider). As such, it will be easy to decommission the old provider when needed.

### DIFF
--- a/shared/agent/src/managers/telemetryManager.ts
+++ b/shared/agent/src/managers/telemetryManager.ts
@@ -9,64 +9,77 @@ import {
 import { Logger } from "../logger";
 import { CodeStreamSession } from "../session";
 import { debug, lsp, lspHandler } from "../system";
-import { TelemetryService } from "../telemetry/telemetry";
-
+import { TelemetryService } from "../telemetry/telemetryService";
+import { SegmentTelemetryService } from "../telemetry/segmentTelemetry";
+import { NewRelicTelemetryService } from "../telemetry/newRelicTelemetry";
 @lsp
 export class TelemetryManager {
-	private readonly _telemetry: TelemetryService;
+	private readonly _providers: TelemetryService[];
 
 	constructor(session: CodeStreamSession) {
-		// TODO: Respect VSCode telemetry opt out
-		this._telemetry = new TelemetryService(session, false);
+		this._providers = [
+			new SegmentTelemetryService(session, false),
+			new NewRelicTelemetryService(session, false),
+		];
 	}
 
 	setConsent(hasConsented: boolean) {
-		this._telemetry.setConsent(hasConsented);
+		this._providers.forEach(provider => provider.setConsent(hasConsented));
 	}
 
 	identify(id: string, props: { [key: string]: any }) {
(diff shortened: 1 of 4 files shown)


shared/agent/src/managers/telemetryManager.ts
```diff
@@ -9,64 +9,77 @@ import {
...
+			} catch (ex) {
+				Logger.error(ex, cc);
+			}
+		});
 	}
 
 	@lspHandler(GetAnonymousIdRequestType)
 	getAnonymousId() {
-		return this._telemetry.getAnonymousId();
+		return this._providers[0].getAnonymousId();
```
CalvinAllen, line 71: Maybe a comment here that explains all providers have the same `anonymousId`, so it doesn't matter which one? Also, any fear of the collection not having any providers?

shared/agent/src/telemetry/newRelicTelemetry.ts
```diff
@@ -0,0 +1,182 @@
...
+		session.ready().then(() => this.initialize());
+
+		const props = {
+			...opts,
+			Endpoint: session.versionInfo.ide.name,
+			"Endpoint Detail": session.versionInfo.ide.detail,
+		};
+		this._superProps = props;
+		this._hasOptedOut = hasOptedOut;
+		this._anonymousId = uuid();
```
CalvinAllen, line 45: Does it matter here, if something gets tracked with this `anonymousId` before the `setAnonymousId` gets called?

shared/agent/src/telemetry/newRelicTelemetry.ts
```diff
@@ -0,0 +1,182 @@
...
+			// replacing spaces with nothing, since spaces are not allowed in New Relic custom event names.
+			// adding an "CodeStream_" prefix, since without, the name of the event becomes a collection name.
+			// that makes it _very_ hard to distinguish from other built-in, New Relic-supplied data collections.
+			NewRelic.recordCustomEvent(`CodeStream_${event.replace(/ /g, "")}`, payload);
+		} catch (ex) {
+			Logger.error(ex, cc);
+		}
+	}
+
+	private flatten(payload: any): any {
```
CalvinAllen, line 159: Is this only good for two levels of object nesting?

CalvinAllen (approved):

shared/agent/src/telemetry/newRelicTelemetry.ts
bcanzanella: yep, it was mostly a bandaid to get things working asap, but eventually anything nested should be changed before it gets to here.

shared/agent/src/managers/telemetryManager.ts
bcanzanella: good call... i'll guard it

shared/agent/src/telemetry/newRelicTelemetry.ts
bcanzanella: should be ok, that uuid is a fallback

dsellarsnr (approved):

PR #1848, opened 2023-11-17
copy node_modules to vs dirs

dsellarsnr:


shared/agent/esbuild.ts
```diff
@@ -83,6 +83,20 @@ const postBuildCopy: CopyStuff[] = [
...
+		to: path.resolve(
+			__dirname,
+			"../../vs/src/CodeStream.VisualStudio.Vsix.x86/agent/node_modules/"
+		),
+	},
+	{
+		from: path.resolve(`${outputDir}/node_modules/**`),
+		to: path.resolve(
+			__dirname,
+			"../../vs/src/CodeStream.VisualStudio.Vsix.x64/agent/node_modules/"
```
CalvinAllen, line 97: Does VS need to do anything with this node_module folder? I.e., does it need bundled in and included in the output?

CalvinAllen (approved):

shared/agent/esbuild.ts
dsellarsnr: yes, needs to be copied as is. has to be in same dir as agent.js

PR #1852, opened 2023-11-22
Add support for span open in IDE

purplexa:


shared/ui/index.tsx
```diff
@@ -659,6 +664,130 @@ function listenForEvents(store) {
...
+										9999
+									),
+									atTop: true,
+								});*/
+							}
+						}
+
+						store.dispatch(
+							setCurrentTransactionSpan({
+								spanId,
```
bcanzanella, line 773: should "1" be removed here? probably don't want that for non-NR folks

shared/agent/src/providers/newrelic.ts
```diff
@@ -3168,6 +3175,215 @@ export class NewRelicProvider
...
+			}
+			`,
+			{
+				accountId: request.accountId,
+			}
+		);
+
+		const partialData = response.actor.account.nrql.results.reduce(
+			(accum, val) => {
+				accum[val.endTimeSeconds * 1000] = accum[val.endTimeSeconds * 1000] || {};
```
eric-j, line 3287: could we assign  `accum[val.endTimeSeconds * 1000]` to a variable to for less calculations and readability?

bcanzanella:
one comment from me, otherwise looks great.

bcanzanella:
just need a final review, but let's wait to merge it.

PR #1856, opened 2023-11-27
* Fix missing node_modules in debug mode

dsellarsnr:
* Always copy node_modules / extra libs even if not macos


dsellarsnr:
Confirmed develop broken on Windows and this PR fixes it

PR #1859, opened 2023-11-28
expanded state should be global, not entityGuid specific for top leve…

eric-j:
…l dropdowns

### DIFF
--- a/shared/ui/Stream/SecurityIssuesWrapper.tsx
+++ b/shared/ui/Stream/SecurityIssuesWrapper.tsx
@@ -7,7 +7,7 @@ import {
 	riskSeverityList,
 	Vuln,
 } from "@codestream/protocols/agent";
-import { isEmpty, lowerCase } from "lodash-es";
+import { isEmpty, lowerCase, isUndefined } from "lodash-es";
 import React, { useEffect, useState } from "react";
 import styled from "styled-components";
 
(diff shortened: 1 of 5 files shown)


shared/ui/Stream/SecurityIssuesWrapper.tsx
```diff
@@ -266,13 +266,14 @@ export const SecurityIssuesWrapper = React.memo((props: Props) => {
...
 
-		const securityIssuesDropdownExpanded =
-			preferences?.securityIssuesDropdownExpanded && props?.entityGuid
-				? preferences.securityIssuesDropdownExpanded[props?.entityGuid]
-				: true;
+		const securityIssuesDropdownIsExpanded = isUndefined(
+			preferences?.securityIssuesDropdownIsExpanded
+		)
+			? true
+			: preferences?.securityIssuesDropdownIsExpanded;
```
dsellarsnr, line 273: It looks like you could these types of expressions simpler using the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing). It is superior to the `||` operator which would indeed fail here. 

```
const securityIssuesDropdownIsExpanded = preferences?.securityIssuesDropdownIsExpanded ?? true;
```

reneelpetit (approved):

PR #1865, opened 2023-12-04
Add full dep list so we don't miss any require at runtime (have seen …

dsellarsnr:
…error logs about missing https-proxy-agent).

cpy is too unstable for this so we use our own copy with globs.


shared/agent/esbuild.ts
```diff
@@ -9,36 +9,154 @@ import { commonEsbuildOptions, processArgs, startEsbuild } from "../build/src/es
...
+
+  fsevents is external because it has a native module
+ */
 const externals = [
 	"fsevents",
-	"undici",
-	"sync-rpc",
-	"sync-request",
-	"newrelic",
+	"@aws-crypto",
```
bcanzanella, line 25: think there's anything the agent team can do for us here? seems like we're taking on some fragility with this?

CalvinAllen (approved):

PR #1903, opened 2024-01-22
warning message for browser errors without a source map uploaded

eric-j:


shared/util/src/protocol/agent/agent.protocol.nr.ts
```diff
@@ -5,6 +5,7 @@ import { NotificationType, RequestType } from "vscode-languageserver-protocol";
 import { Project, RepoProjectType } from "./agent.protocol.scm";
 import { CSRepository, CSStackTraceInfo, CSStackTraceLine } from "./api.protocol.models";
 import { Uri } from "vscode";
+import { JsxAttribute } from "typescript";
```
dsellarsnr, line 8: accidental import?

eric-j: Yep, removed

PR #1911, opened 2024-01-23
Rethrow exceptions after max retries

CalvinAllen:
Reason: Now that we have places where actual NRQL queries can have user input, it is possible to input a bad query. Previously, even though `query` had a retry limit, the inner portions of `clientRequestWrap` weren't throwing and were in a traditional `while` loop that would never end.

1. Pass retries from `query` down to `clientRequestWrap`
2. Make `clientRequestWrap` only retry the specified number of times
3. Every time through the loop; capture the exception
4. If we have no response but do have an exception at the end, throw it.


dsellarsnr (approved):
+1 say to to infinity

bcanzanella (approved):

PR #1919, opened 2024-01-30
Fix retry loop

dsellarsnr:
* Handle case where refresh token invalid - logout user and return to login screen

### DIFF
--- a/jb/src/main/kotlin/agent/CodeStreamLanguageClient.kt
+++ b/jb/src/main/kotlin/agent/CodeStreamLanguageClient.kt
@@ -136,6 +136,13 @@ class CodeStreamLanguageClient(private val project: Project) : LanguageClient {
     @JsonNotification("codestream/didFailLogin")
     fun didFailLogin(json: JsonElement?) {}
 
+    @JsonNotification("codestream/didEncounterInvalidRefreshToken")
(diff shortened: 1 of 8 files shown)


marcelofarias (approved):

jb/src/main/kotlin/agent/CodeStreamLanguageClient.kt
```diff
@@ -136,6 +136,13 @@ class CodeStreamLanguageClient(private val project: Project) : LanguageClient {
     @JsonNotification("codestream/didFailLogin")
     fun didFailLogin(json: JsonElement?) {}
 
+    @JsonNotification("codestream/didEncounterInvalidRefreshToken")
```
bcanzanella, line 139: do we need also need this for VS?

dsellarsnr: yes, meant to ask @CalvinAllen

dsellarsnr: should we merge this and to VS in a separate PR?

bcanzanella: @dsellarsnr yes

PR #1928, opened 2024-02-05
Improve nrai access control checking

purplexa:


shared/ui/Stream/CodeError/index.tsx
```diff
@@ -1912,6 +1913,10 @@ const CodeErrorForCodeError = (props: PropsWithCodeError) => {
 		}
 	}
 
+	useDidMount(() => {
+		dispatch(getNrCapability("nrai"));
```
dsellarsnr, line 1917: There is a probably a race condition here where getNrCapability starts, rendering starts with grok flag off, then getNrCapability resolves and grok flag toggles on mid-render. This _might_ be OK but it's hard to validate.

purplexa: Yeah, I was aware, but I couldn't find a better way to resolve it. In Observability.tsx there's a call to bootstrapNrCapabilities() that *should* result in the data getting fetched before we get here, but I didn't want to totally rely on it.

reneelpetit (approved):

PR #1929, opened 2024-02-05
Feature/renaming events clean up

reneelpetit:
https://new-relic.atlassian.net/browse/NR-220932
https://new-relic.atlassian.net/browse/NR-220043
https://new-relic.atlassian.net/browse/NR-220030

### DIFF
--- a/shared/agent/src/managers/postsManager.ts
+++ b/shared/agent/src/managers/postsManager.ts
@@ -520,27 +520,38 @@ function trackPostCreation(
 						}
 					}
 
+					let entryPoint = "global_nav";
+					if (request.entryPoint != null) {
+						switch (request.entryPoint) {
+							case "Gutter":
+								entryPoint = "gutter";
+								break;
+							case "Global Nav":
+								entryPoint = "global_nav";
+								break;
+							case "Shortcut":
+								entryPoint = "shortcut";
+								break;
+							case "Lightbulb Menu":
+								entryPoint = "lightbulb_menu";
+								break;
+							case "Action List":
+								entryPoint = "action_list";
+								break;
+							case "Hover Icons":
+								entryPoint = "hover_icons";
+								break;
+							case "Advanced Link":
+								entryPoint = "advanced_link";
+								break;
+						}
+					}
+
 					if (request.codemark) {
 						// this is a standard codemark -- note its event name includes "created" rather than "reply"
 						const { markers = [] } = request.codemark;
 						const codemarkProperties: TelemetryData = {
-							meta_data: `entry_point: ${
-								request.entryPoint === "Gutter"
-									? "gutter"
-									: request.entryPoint === "Global Nav"
-									? "global_nav"
-									: request.entryPoint === "Shortcut"
-									? "shortcut"
-									: request.entryPoint === "Lightbulb Menu"
-									? "lightbulb_menu"
-									: request.entryPoint === "Action List"
-									? "action_list"
-									: request.entryPoint === "Hover Icons"
-									? "hover_icons"
-									: request.entryPoint === "Advanced Link"
-									? "advanced_link"
-									: ""
-							}`,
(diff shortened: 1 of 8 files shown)


shared/agent/src/managers/postsManager.ts
```diff
@@ -527,22 +527,25 @@ function trackPostCreation(
...
 									? "lightbulb_menu"
-									: "Action List"
+									: request.entryPoint === "Action List"
 									? "action_list"
-									: "Hover Icons"
+									: request.entryPoint === "Hover Icons"
 									? "hover_icons"
-									: "Advanced Link"
+									: request.entryPoint === "Advanced Link"
 									? "advanced_link"
```
dsellarsnr, line 541: Probably a good time to cleanup this code for clarity. 

Option 1 is to use a case statement and lift it outside the string interpolation (define an entryPoint variable before this line). 

Option 2 is to define a map near the top of the file 
```
const entryPointMap: Record<string, string> = {
	"Gutter": "gutter",
	"Global Nav": "global_nav",
	//...
}
```

and then lookup the value from the map

```
const entryPoint = entryPointMap[request.entryPoint] ?? "";
```

and option 3 since it follows a convention would be to use lodash [snakeCase](https://lodash.com/docs/4.17.15#snakeCase)
```
const entryPoint = snakeCase(request.entryPoint)
```

bcanzanella: +1

shared/ui/Stream/MethodLevelTelemetry/ObservabilityAnomalyPanel.tsx
```diff
@@ -322,7 +322,21 @@ export const ObservabilityAnomalyPanel = () => {
 					<span style={{ marginRight: "6px" }}>
 						{derivedState.currentObservabilityAnomaly.name}
 					</span>
-					{titleHovered && <Icon title="Open on New Relic" delay={1} name="link-external" />}
+					{titleHovered && (
```
bcanzanella, line 325: the couple of onClicks w/icons don't seem correct to me... we usually don't attach click handlers to them -- and in this case, we should probably attach to the <a> tag instead. i'll ping u.

shared/agent/src/managers/postsManager.ts
reneelpetit: ok thanks!  fixed

shared/ui/Stream/MethodLevelTelemetry/ObservabilityAnomalyPanel.tsx
reneelpetit: ok I think I got them, thanks!

shared/ui/Stream/CodeError/index.tsx
```diff
@@ -929,10 +929,21 @@ export const BaseCodeErrorHeader = (props: PropsWithChildren<BaseCodeErrorHeader
 									{props.errorGroup?.errorGroupUrl && props.codeError.title ? (
 										<span data-testid="code-error-title">
 											<Link
-												href={
-													props.errorGroup.errorGroupUrl! +
-													`&utm_source=codestream&utm_medium=ide-${derivedState.ideName}&utm_campaign=error_group_link`
-												}
+												onClick={e => {
+													HostApi.instance.track("codestream/link_to_newrelic clicked", {
```
bcanzanella, line 933: this needs a `e.preventDefault();` but please test and ensure it works after adding.

bcanzanella (changes requested):

shared/ui/Stream/CodeError/index.tsx
reneelpetit: added preventDefault and double-checked them all again to make sure they work. They seem to work for me

bcanzanella (approved):

PR #1934, opened 2024-02-06
visualziation selection for query builder

eric-j:
https://new-relic.atlassian.net/browse/NR-218056

### DIFF
--- a/shared/agent/src/providers/newrelic/nrql/nrqlProvider.ts
+++ b/shared/agent/src/providers/newrelic/nrql/nrqlProvider.ts
@@ -319,10 +319,10 @@ export class NrNRQLProvider {
 						: "json";
 				return {
 					selected: selectedValue,
-					enabled: selectedValue === "billboard" ? ALL_RESULT_TYPES : ["json"],
+					enabled: ["json"],
 				};
 			} else {
-				return { selected: "billboard", enabled: ALL_RESULT_TYPES };
+				return { selected: "billboard", enabled: ["json"] };
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


shared/agent/src/providers/newrelic/nrql/nrqlProvider.ts
```diff
@@ -302,16 +307,22 @@ export class NrNRQLProvider {
...
+					Object.keys(value).length === 1 && !Array.isArray(value[Object.keys(value)[0]])
+						? "billboard"
+						: "json";
+				return {
+					selected: selectedValue,
+					enabled: selectedValue === "billboard" ? ALL_RESULT_TYPES : ["json"],
+				};
 			} else {
-				return "billboard";
+				return { selected: "billboard", enabled: ALL_RESULT_TYPES };
```
bcanzanella, line 325: technically this works, though not terribly useful. on NR1 billboard's only other alternatives are ["json"] (and bullet, which we dont support)

shared/ui/Stream/Dropdown.tsx
```diff
@@ -71,15 +71,22 @@ export const Dropdown = (props: Props) => {
 		props.selectedValue || ""
 	);
 
+	useEffect(() => {
```
bcanzanella, line 74: there any risk[s] in modifying this component? i ask because it's used in a lot of places.

bcanzanella:
couple ❓ s otherwise LGTM

shared/agent/src/providers/newrelic/nrql/nrqlProvider.ts
eric-j: good point, ill change

shared/ui/Stream/Dropdown.tsx
eric-j: Low risk, and this actually fixes some wonky behavior where its currently implemented as well.

bcanzanella: ok sweet

bcanzanella (approved):

PR #1940, opened 2024-02-07
What's New

CalvinAllen:
* WhatsNew.json ( can be an array of objects with version: string, title: string) allows triggering notifications on an exact match release, with customizable title (deployed into /dist) - Agent does all the work
* Profile -> Help -> What's New for triggering at will


shared/agent/src/session.ts
```diff
@@ -705,6 +707,31 @@ export class CodeStreamSession {
...
+
+		const isFlagged = whatsNew.includes(currentVersion);
+
+		if (isFlagged && preferences) {
+			if (!preferences.whatsNewSeen?.includes(currentVersion)) {
+				this.agent.sendNotification(WhatsNewNotificationType, {
+					title: `CodeStream has been upgraded to v${currentVersion}!`,
+				});
+				preferences.whatsNewSeen?.push(currentVersion);
+				this._api?.updatePreferences({ preferences });
```
dsellarsnr: This should probably be in ui in an useDidMount - you can have the codestream extension hidden and never open it up but still be logged in and reach this code. On next login when you open the webview you won't get the whatsnew.

CalvinAllen: @dsellarsnr Not sure I'm following - this is only to trigger an IDE notification, so even if the UI isn't visible and they get logged in, they'd still get the notification which gives them access to the whats new. Once they get that notification, we consider it "seen".

CalvinAllen: @dsellarsnr Maybe its the preference name giving way to confusion here...it doesn't mean they actually SAW the what's new, it just means they've seen the notification

dsellarsnr: here is the step by step

- open vscode with codestream collapsed
- agent sends whatsnew notification, mark preferences whatsNewSeen
- close vscode
- open vscode, open codestream extension, no whatsnew notification / whatsnew UI sown

shared/agent/src/session.ts
```diff
@@ -705,6 +707,40 @@ export class CodeStreamSession {
 		return data[0];
 	}
 
+	@log()
+	async whatsNewNotification() {
+		try {
+			const me = await SessionContainer.instance().users.getMe();
+			const preferences = me.preferences;
+
+			const whatsNewBuffer = fs.readFileSync(path.join(__dirname, "WhatsNew.json"), {
```
bcanzanella, line 716: you could bail out early here and not read from the fileSystem if preferences.whatsNewSeen[this.versionInfo.extension.version] = true.

also to the naming convo you had with Dillon, maybe whatsNewNotifications or whatsNewNotificationsSent is closer to what is being tracked?

shared/agent/src/session.ts
CalvinAllen: @dsellarsnr At the point the agent sends the notification, the IDE is responsible for showing a native notification. That's all that needs to be seen.  Once that notification has triggered (which it does even when the CS panel is collapsed, assuming auto-login), its done and considered "seen".

shared/agent/src/session.ts
CalvinAllen: Done

shared/ui/Stream/WhatsNew.tsx
```diff
@@ -0,0 +1,244 @@
...
+						transform: "scale(2)",
+						display: "inline-block",
+						marginRight: "15px",
+						top: "15px",
+					}}
+					name="newrelic"
+				/>
+				What's New
+			</h1>
+			<TimelineContent>
```
bcanzanella, line 77: @CalvinAllen  is this component going to be the source of truth for the content or the WhatsNew.json?

CalvinAllen: @bcanzanella You could create a whats new entry for a point release, but not a notification, for instance. Or, do both for a major.

bcanzanella: ah gotcha, maybe nuke the content in that fake component and add the latest from 15.2 just so there's real content?

```

## [15.2.0] - 2023-2-7

- A new "Apply Fix" button allows you to easily accept a suggested code fix when NRAI analyzes an error for you
- Code fixes suggested by NRAI are now presented in a diff view
- 
-```

File: shared/agent/src/session.ts [hunk 4]
```diff
@@ -705,6 +707,45 @@ export class CodeStreamSession {
...
+
+			// already tracked for this version; bail out
+			if (preferences?.whatsNewNotificationsSent?.includes(currentVersion)) {
+				return;
+			}
+
+			const whatsNewBuffer = fs.readFileSync(path.join(__dirname, "WhatsNew.json"), {
+				encoding: "utf-8",
+			});
+			const whatsNew: { version: string; title: string }[] = JSON.parse(whatsNewBuffer);
```
bcanzanella, line 725: would there ever be > 1 message for a version? should the format be an {} instead of array with version the key? also does would that affect VS with its diff version syntax?

shared/agent/src/session.ts
dsellarsnr: Got it, should be good then. I tried to see a whats new, got it to trigger all the way to window.showInformationMessage in vscode but didn't see anything 🤨

shared/agent/src/session.ts
CalvinAllen: @bcanzanella I wouldn't think so, but I guess that's a question for Dave? At least in VSC, the notification can only be a single string / "title", and you can't have multiple blocks/paragraphs/etc, and it doesn't support line breaks or anything.

shared/ui/Stream/WhatsNew.tsx
CalvinAllen: @bcanzanella Ah yeah, I can do that - that was mostly in there to figure out the styling and what not, but now that we have real notes, I'll update it.

shared/ui/Stream/EllipsisMenu.tsx
```diff
@@ -363,6 +367,18 @@ export function EllipsisMenu(props: EllipsisMenuProps) {
...
 			submenu: [
+				{
+					label: "What's New",
+					key: "whatsnew",
+					action: () => {
+						HostApi.instance.notify(OpenEditorViewNotificationType, {
+							panel: "whatsnew",
+							title: "What's New",
+							entryPoint: "profile",
+							entityAccounts: [],
```
bcanzanella, line 378: not a fix for this PR, but maybe we come up with a different object model for OpenEditorViewNotificationType where we dont have to stuff everything OpenEditorViewNotification. food for thought.

CalvinAllen: Yeah, its already starting to show its "teeth", but now that we have a few more examples, hopefully it'll be easier to reason about.

shared/agent/src/session.ts
bcanzanella: prolly fine with the "title" / string stuff then, maybe just check that the version in that versionInfo file for VS wont be an issue?

bcanzanella: even though VS doesnt support toasts, in the future we could do something like highlighting other parts of VS chrome or something inside the sidebar itself to draw attention

shared/ui/Stream/WhatsNew.tsx
CalvinAllen: @bcanzanella Added priors from 15.0.0

shared/agent/src/session.ts
CalvinAllen: I've been pushing a few changes; there were a couple issues after I rebased.

PR #1954, opened 2024-02-12
Whats New Notification for JB

CalvinAllen:
Triggers notification on login, and opens editor view

Notes:

1. Copied some of the logic from the WebViewRouter into the notification component so that the sidebar doesn't have to be visible for this to work.
2. If the notification triggers and you "accept" _before_ IntelliJ/etc. have finished indexing, the panel shows an error.

### DIFF
--- a/jb/src/main/kotlin/notification/NotificationComponent.kt
+++ b/jb/src/main/kotlin/notification/NotificationComponent.kt
@@ -5,25 +5,31 @@ import com.codestream.agentService
 import com.codestream.appDispatcher
 import com.codestream.clmService
 import com.codestream.codeStream
+import com.codestream.gson
 import com.codestream.protocols.agent.Codemark
-import com.codestream.protocols.agent.FollowReviewParams
 import com.codestream.protocols.agent.ObservabilityAnomaly
 import com.codestream.protocols.agent.Post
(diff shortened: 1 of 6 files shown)


jb/src/main/kotlin/notification/NotificationComponent.kt
```diff
@@ -107,6 +113,25 @@ class NotificationComponent(val project: Project) {
...
+        notification.addAction(NotificationAction.createSimple("See What's New") {
+            ApplicationManager.getApplication().invokeLater {
+                val editorManager = FileEditorManager.getInstance(project)
+                val editorNotification = EditorOpenNotification("What's New", "whatsnew", null, null)
+                val file = WebViewEditorFile.create(gson.toJsonTree(editorNotification))
+
+                val editor = editorManager.openFile(file, true, true).firstOrNull()
+                editor?.component?.repaint()
+
+                notification.expire()
```
dsellarsnr, line 127: Would this close the notification right away?

CalvinAllen: It closes it when you click the action ("See What's New"), otherwise it stayed open

dsellarsnr (approved):

PR #1956, opened 2024-02-13
codify promotion link for nrql files

eric-j:


reneelpetit (approved):

shared/ui/Stream/NRQL/NRQLPanel.tsx
```diff
@@ -101,6 +101,11 @@ const ResizeEditorContainer = styled.div`
 	padding: 8px;
 `;
 
+const CodeText = styled.span`
+	font-family: "Courier New", Courier, monospace;
```
purplexa, line 105: I think elsewhere we use `font-family: Menlo, Consolas, "DejaVu Sans Mono", monospace;`

eric-j: good call

PR #1967, opened 2024-02-14
fix indenting whats new page

eric-j:
@planteater The indent was because the paragraphs were wrapped in `<ListContainer>`, you only want to use that if you are going to make a bulleted list.

I removed them here, as well as increased the margin between a title like "Log Search" and the below paragraph.


eric-j:
image

planteater (approved):

PR #1974, opened 2024-02-19
nrql updates

bcanzanella:
- Wrong accountId is used in certain cases - NR-230636
- NRQL queries with comments fails - NR-229045
-- additional grammar cleanup w/r/t comments
- Don't assume a "facet" property - NR-229519
-- adds support for multi-faceted bar + pie charts
- Fixes Agent "Run UnitTests for this File"
- Adds ability to convert monaco editor into a simple textarea.

### DIFF
--- a/shared/agent/test/unit/providers/newrelic/nrqlDocumentParser.spec.ts
+++ b/shared/agent/test/unit/providers/newrelic/nrqlDocumentParser.spec.ts
@@ -115,4 +115,42 @@ SELECT foo from Log where appName = 'bar'
 			},
 		]);
 	});
+
+	it("should handle embedded comments", () => {
+		const content = `FROM Transaction
+SELECT name /* that's the bar
+on a new line */
+WHERE name is not null // baz is here`;
+		const results = new NrqlDocumentParser().parse(content);
+		expect(results.length).toStrictEqual(1);
+		expect(results).toStrictEqual([
+			{
+				invalid: false,
+				text: content,
+				range: {
+					start: 0,
+					end: 100,
+				},
+			},
+		]);
+	});
+
+	it("should handle embedded comments", () => {
+		const content = `/*
+FROM Transaction SELECT name WHERE name is not null 
+*/`;
+		const results = new NrqlDocumentParser().parse(content);
+		expect(results.length).toStrictEqual(1);
+		expect(results).toStrictEqual([
+			{
+				invalid: false,
+				text: `FROM Transaction SELECT name WHERE name is not null 
+`,
+				range: {
+					start: 3,
+					end: 55,
+				},
+			},
+		]);
+	});
 });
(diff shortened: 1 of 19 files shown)


shared/agent/test/unit/providers/newrelic/nrqlDocumentParser.spec.ts
CalvinAllen, line 1: 🔥

shared/agent/test/unit/providers/newrelic/nrqlProvider.spec.ts
CalvinAllen, line 1: 🔥

shared/ui/Stream/NRQL/NRQLEditor.tsx
```diff
@@ -21,20 +24,58 @@ export const NRQLEditor = React.forwardRef(
 			onSubmit?: (e: { value: string | undefined }) => void;
 			setValue?: (e: { value: string | undefined }) => void;
 			isReadonly?: boolean;
+			// if false, editor will fallback to a simple <textarea>, true & undefined are the same
```
CalvinAllen, line 27: `undefined` being equivalent to `true` seems kind of odd in my head. Anyway to get rid of `undefined`, but default to `true` so its always explicit?

shared/ui/Stream/NRQL/NRQLResultsLine.tsx
```diff
@@ -15,7 +15,16 @@ const formatXAxisTime = time => {
 	return new Date(time).toLocaleTimeString();
 };
 
-export const NRQLResultsLine = (props: { results: NRQLResult[] }) => {
+interface Props {
```
CalvinAllen, line 18: No way to reuse this prop type across all the facet types?

CalvinAllen (approved):
Some questions, but nothing I'll die on a hill about 😀

shared/ui/Stream/NRQL/NRQLEditor.tsx
bcanzanella: maybe it should be flipped so that the non-enhanced version has to be explicitly stated? useNonEnhancedEditor? useSimpleEditor?

shared/ui/Stream/NRQL/NRQLResultsLine.tsx
bcanzanella: probably not without introducing some generics for the properties that are specific to each chart type?  inheritance could get kinda messy... something like this:

```
interface NRQLResultsLineProps {	 
	facet: string[];
}

interface NRQLResultsProps<T> {
	results: NRQLResult[];
	options: NRQLResultsLineProps;
}

export const NRQLResultsLine = (props: NRQLResultsProps<NRQLResultsLineProps>) {}
```

shared/ui/Stream/NRQL/NRQLEditor.tsx
CalvinAllen: If going that route, I don't like negative booleans, so much prefer the positive version of `useSimpleEditor` 👍🏻

shared/ui/Stream/NRQL/NRQLResultsLine.tsx
CalvinAllen: Eh, not worth it. Just glancing they all seemed to be the same, but I didn't detail them all out.

shared/ui/Stream/NRQL/NRQLEditor.tsx
bcanzanella: pushed that change

CalvinAllen (approved):
🚀

PR #1979, opened 2024-02-21
NR-220824 Solidify the dev experience related to watchers, pre-commit…

dsellarsnr:
… hooks, and CI tests

- Bump esbuild
- Make agent build fast again with symlink in watch mode
- Fix incremental config to store .tsbuildinfo be per-project root instead of in repo root
- Log hook execution for future troubleshooting


dsellarsnr:
Wasn't able to reproduce any hook compile errors so I added logging to `.husky/_/pre-commit.log` so we can see if hooks are not running at all sometimes or if they are running and compilation passes when it shouldn't.

jb/tsconfig.json
```diff
@@ -22,6 +23,7 @@
 		"sourceMap": true,
 		"strict": true,
 		"target": "es2017",
+		"tsBuildInfoFile": ".tsbuildinfo",
```
bcanzanella, line 26: is this a new file?

bcanzanella: nm, it's at the proj root -- i had the workspace open

bcanzanella (approved):

jb/tsconfig.json
dsellarsnr: It was in the root tsconfig.json before but that turns out to be wrong for inheriting the config. It was creating `.tsbuildinfo` in the root but it only applied to the last `tsc` run, so running agent compile would create an agent specific `.tsbuildinfo` and then running ui `tsc` would reject the agent specific `.tsbuildinfo` and overwrite it. 

By putting the `incremental` and `tsBuildInfoFile` setting in each sub-project the `.tsbuildinfo` is no longer in the root and it's now in the right place in the sub-project

bcanzanella: ah gotcha

PR #1980, opened 2024-02-21
NR-223660: Creating an issue on Shortcut results in two duplicte issues being created on Shortcut

reneelpetit:


dsellarsnr (approved):
Yup def needs to go

PR #1994, opened 2024-02-28
Moar Token Shtuff

CalvinAllen:
The two various notifications were battling against each other in all out gorilla warfare. Remove the (albeit, Original) notification and utilize the one that triggers the UX mechanism for logging out and navigating to the login screen


CalvinAllen:
Part of the issue was branch 1 being opened for an extended period, while branch 2 was opened and closed before branch 1 ever made it to completion. Once they both came together, it was a no-go and caused various issues.

cstryker:
What a fustercluck.

PR #1997, opened 2024-02-29
Feature/nr 111940 nrai demo mode

dsellarsnr:


shared/ui/Stream/CodeErrorNav.tsx
```diff
@@ -602,18 +606,18 @@ export function CodeErrorNav(props: Props) {
 					await dispatch(
 						updateCodeError({
 							...derivedState.codeError,
-							accountId: errorGroupResult.accountId,
-							title: errorGroupResult.errorGroup?.title || "",
-							text: errorGroupResult.errorGroup?.message || undefined,
+							// accountId: errorGroupResult.accountId,
```
bcanzanella, line 609: these commented out are ok? if so, perhaps just remove them all?

shared/ui/src/components/Pane.tsx
```diff
@@ -233,14 +246,44 @@ export const PaneHeader = React.memo((props: PropsWithChildren<PaneHeaderProps>)
...
+			dispatch(setDemoMode(nextDemoMode));
+			captured = "";
+		}
+	};
+
 	if (props.noDropdown) {
 		return (
 			<PaneHeaderRoot
+				onClick={demoClick}
+				onContextMenu={demoClick}
```
bcanzanella, line 278: does this need to be guarded at all? aka will non-demo folks be able to trigger this?

shared/ui/store/codeErrors/api/CodeErrorsApi.ts
```diff
@@ -0,0 +1,54 @@
+import {
```
bcanzanella, line 1: nice!

bcanzanella (changes requested):
just a couple comments/questions

shared/ui/src/components/Pane.tsx
dsellarsnr: i made it a cheat code, you have to mouse click left, right, left, right, left, left

shared/ui/Stream/CodeErrorNav.tsx
dsellarsnr: the server call is pretty close to matching the type - only allows updates to title, text, and stackTraces fields so `objectInfo` update never did anything here.

bcanzanella (approved):

PR #1999, opened 2024-02-29
entity guid finder

bcanzanella:
Screenshot 2024-02-29 at 3 00 02 PM
- adds editor decoration to NR entity guids with associated metadata. hover links to NRQL using first goldenMetric query for an entity 
- nrql-click on hover's telemetry as "entity_guid_finder" [Dave?]
- only in VSCode
- entity guids are found on active editor change and on document save (not as you type)
- they can be found in any file type
- agent caching layer for entityGuids (should that be expanded for all Entity lookups?)
- enhanced parseId with rules from the spec, https://github.com/newrelic/entity-definitions/blob/main/docs/entities/guid_spec.md , defaulted to non-strict mode for existing usage; renamed properties


bcanzanella:
ready for review, but not going to merge.

shared/agent/src/providers/newrelic/utils.ts
```diff
@@ -22,25 +22,61 @@ export function mapNRErrorResponse(ex: Error): NRErrorResponse {
...
+	[key: string]: RegExp;
+};
+
+/**
+ * Parses an ID-like string and returns the corresponding NewRelicId object.
+ * @param idLike - ID-like string to be parsed.
+ * @param strict - When set to true, performs strict validation on the parsed values. Default is false.
+ * @returns The parsed NewRelicId object or undefined if parsing fails or the ID-like string is empty.
+ */
+export function parseId(idLike: string, strict: boolean = false): NewRelicId | undefined {
```
CalvinAllen, line 35: I think there is a version of this in the UI layer, too, might be worth checking there and syncing them

CalvinAllen:
Preemptive 🔥

shared/agent/src/providers/newrelic/utils.ts
bcanzanella: i'll add it

bcanzanella:
@CalvinAllen just wanted to make sure the stuff i did w/r/t the Log panel looked ok. wanted to add that entry point from the hover to Logs, but had to plumb in a "fake" EntityAccount to appease the logic

CalvinAllen:
> @CalvinAllen just wanted to make sure the stuff i did w/r/t the Log panel looked ok. wanted to add that entry point from the hover to Logs, but had to plumb in a "fake" EntityAccount to appease the logic

How come you couldn't just pass in the EntityGuid itself?

shared/agent/src/providers/newrelic/utils.ts
bcanzanella: pushed

CalvinAllen:
> > > @CalvinAllen just wanted to make sure the stuff i did w/r/t the Log panel looked ok. wanted to add that entry point from the hover to Logs, but had to plumb in a "fake" EntityAccount to appease the logic
> > 
> > 
> > How come you couldn't just pass in the EntityGuid itself?
> 
> from what i saw, the request to the agent in the APMLog panel that populated the EntityAccount is only related to repos open in the IDE and these entityGuids can be from anything.

Ah, right, right. Makes sense then, didn't realize it could be anything.

bcanzanella:
> > > > @CalvinAllen just wanted to make sure the stuff i did w/r/t the Log panel looked ok. wanted to add that entry point from the hover to Logs, but had to plumb in a "fake" EntityAccount to appease the logic
> > > 
> > > 
> > > How come you couldn't just pass in the EntityGuid itself?
> > 
> > 
> > from what i saw, the request to the agent in the APMLog panel that populated the EntityAccount is only related to repos open in the IDE and these entityGuids can be from anything.
> 
> Ah, right, right. Makes sense then, didn't realize it could be anything.
> 
> Should we expand the logic in the component to be more "wide open" for services, so you don't have to fake a bunch of stuff to make it work?

that was kind of my idea with what i added, but if you think there's a cleaner way, i'm for it. `GetObservabilityReposRequestType` is probably the biggest limiting factor.

bcanzanella:
@CalvinAllen i rebased to get Lexa's changes incorporated

shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
```diff
@@ -285,6 +283,7 @@ export const APMLogSearchPanel = (props: {
 	}) => {
 		if (!entityAccount) {
 			setSelectedEntityAccount(null);
+			return;
```
bcanzanella, line 286: this seemed missing to me...

CalvinAllen: Yeah, seems like it. We probably just haven't hit that yet since we have always been able to find an EntityAccount up until now.

CalvinAllen (approved):
:rocket:

PR #2005, opened 2024-03-04
AD - PHP Scopes

CalvinAllen:
"scope" can have double slashes that need escaped. Catch any exceptions at the query level

Example:

```
scope = 'WebTransaction/Custom/App\\Http\\Controllers\\Api\\SomeController@SomeAction'
```

### DIFF
--- a/shared/agent/src/providers/newrelic/anomalyDetectionDrillDown.ts
+++ b/shared/agent/src/providers/newrelic/anomalyDetectionDrillDown.ts
@@ -581,12 +581,7 @@ export class AnomalyDetectorDrillDown {
 	}
 
 	private runNrql<T>(nrql: string): Promise<T[]> {
-		try {
-			return this.graphqlClient.runNrql(this._accountId, nrql, 400);
-		} catch (ex) {
-			Logger.error(ex);
-			return Promise.resolve([]);
-		}
+		return this.graphqlClient.runNrql(this._accountId, nrql, 400);
 	}
 
 	private durationComparisonToAnomaly(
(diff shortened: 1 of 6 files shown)


bcanzanella:
let's get a couple tests in here just to be safe.

CalvinAllen:
@bcanzanella I got a few in there, specifically around escaping the NRQL at the lowest level the scope is used.

bcanzanella (approved):

bcanzanella:
any thoughts @dsellarsnr ?

shared/agent/src/providers/newrelic/anomalyDetectionDrillDown.ts
```diff
@@ -580,7 +581,12 @@ export class AnomalyDetectorDrillDown {
 	}
 
 	private runNrql<T>(nrql: string): Promise<T[]> {
-		return this.graphqlClient.runNrql(this._accountId, nrql, 400);
+		try {
+			return this.graphqlClient.runNrql(this._accountId, nrql, 400);
+		} catch (ex) {
+			Logger.error(ex);
+			return Promise.resolve([]);
+		}
```
dsellarsnr, line 589: The caller should choose to ignore errors or not - APIs should always return errors. If you do it here someone a year for now will have a very bad time figuring out why something silently fails. (logging an error and returning the error is a common pattern so the logging of the error won't necessarily make it stand out)

dsellarsnr:
Just the one comment, the rest looks good

shared/agent/src/providers/newrelic/anomalyDetectionDrillDown.ts
CalvinAllen: @dsellarsnr  I might take this out, because following up the chain, it seems there are plenty of try/catches around the topmost calls anyway. In those cases, I'm not entirely certain anymore this is what's causing the blank CS pane

PR #2009, opened 2024-03-05
multipleRepos flag should be set based on relatedRepos meta call

eric-j:
https://new-relic.atlassian.net/browse/NR-238650

### DIFF
--- a/shared/ui/index.tsx
+++ b/shared/ui/index.tsx
@@ -639,12 +639,6 @@ function listenForEvents(store) {
 						definedQuery.query.occurrenceId =
 							definedQuery.query.occurrenceId || definedQuery.query.traceId;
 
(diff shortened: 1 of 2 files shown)


shared/ui/index.tsx
```diff
@@ -680,6 +674,8 @@ function listenForEvents(store) {
 							{ entityGuid: definedQuery.query.entityId || "" }
 						)) as GetObservabilityErrorGroupMetadataResponse;
 
+						definedQuery.query.multipleRepos = response?.relatedRepos?.length > 1;
```
purplexa, line 677: won't this throw an error if `response.relatedRepos` is undefined since you try comparing undefined to 1? should it be something like
```typescript
definedQuery.query.multipleRepos = !!response?.relatedRepos?.length && response?.relatedRepos?.length > 1;
```

PR #2014, opened 2024-03-06
Support protocol handling to open logs webview

purplexa:


CalvinAllen:
Looks good to me!

I don't quite understand why there isn't matching Entity already in that result set to use, but we're discussing opening that up to _ALL_ entities anyway, so that might simplify that in the long run.

CalvinAllen:
Oops....

CalvinAllen (approved):

bcanzanella (approved):
nice!

dsellarsnr (approved):

PR #2015, opened 2024-03-06
.graphql files w/NR

bcanzanella:
testing out using graphql queries in their own files. runNrql is used in a ton of spots (as a harness) that we'd be able to reuse a lot.


dsellarsnr:
We can go one level deeper and use codgen to have it typed. 

https://github.com/dotansimha/graphql-code-generator

bcanzanella:
> We can go one level deeper and use codgen to have it typed.
> 
> https://github.com/dotansimha/graphql-code-generator

maybe a topic for the friday hackathon??

PR #2018, opened 2024-03-07
adds X-Query-Source-* headers

bcanzanella:
...also cleans up some usages of some urls.


shared/agent/src/providers/newrelic/newrelicHttpClient.ts
```diff
@@ -1,31 +0,0 @@
-import url from "url";
```
bcanzanella, line 1: didn't see any usages of this...

PR #2019, opened 2024-03-07
Visual Studio - Editor Webview

CalvinAllen:
There is A LOT going on here, so let me try and summarize -

Refactoring -
1. Moved the 'editor' (`shared/webviews/editor`) files into the UI folder (`shared/ui`).
2. Renamed `index.tsx` in `shared/ui` to `sidebar.tsx` and added the `editor.tsx` from above
3. Modified VSCode and JetBrains esbuilds to recognize the entrypoint filename change for both files
4. Modified various imports, tsconfig, etc., to handle it
5. In VSCode, renamed the template files from `src/webviews/editor/index` (and etc) to `src/webviews/editor/editor.ts` (and etc)
6. Same for JetBrains, however, I also modified the startup routine to copy the files around based on the new output names.
7. Deleted the entire `shared/webviews` folder - buh bye.

I was able to verify VSCode was fully functional with the changes, but not JB **_so somebody needs to test that locally for me, please and thank you._**

Visual Studio -
1. Refactored the esbuild stuff to handle the webviews - including using a symlink for the node_modules instead of copying them again.
2. Refactored the x86 and x64 project files (by hand!) to use glob patterns for bringing in as many of the files as possible, without actually adding each one, one by one. to the project. 
3. Renamed various files DotNetBrowser -> SidebarBrowser, etc., just in preparation for having _two_ sets of things. WebViewRouter -> SidebarRouter, etc.
4. Moved the webview template files out of the Shared project and into `src/resources` just because I like that better.

I was able to verify that VS was fully functional with the changes (obviously only the sidebar is used at this juncture), but the files are prepped and ready to be used for the editor when ready.


bcanzanella:
@CalvinAllen i can check JB, but will wait until conflicts are resolved.

CalvinAllen:
Go time

vs/src/CodeStream.VisualStudio.Core/SolutionInfo.cs
```diff
@@ -1,15 +1,15 @@
...
-	}
-}
+namespace CodeStream.VisualStudio.Core.Properties
+{
+	public static class SolutionInfo
+	{
+		/// <summary>
+		/// This was generated by a tool (build process)
+		/// </summary>
+		public const string Version = "15.4.1.0";
```
bcanzanella, line 8: should this be 15.4.2.0? (for the next release)

vs/src/resources/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+[assembly: AssemblyCompany("New Relic, Inc.")]
+[assembly: AssemblyProduct("CodeStream")]
+
+[assembly: ComVisible(false)]
+
+[assembly: AssemblyVersion("15.4.1.0")]
```
bcanzanella, line 9: 15.4.2.0 ?

bcanzanella (changes requested):
Screenshot 2024-03-08 at 11 06 54 AM

JB died, hangs at a blank sidebar panel. prod JB ext is good

CalvinAllen:
> Screenshot 2024-03-08 at 11 06 54 AM
> JB died, hangs at a blank sidebar panel. prod JB ext is good

@bcanzanella Thanks, can you try again?

bcanzanella (approved):

PR #2021, opened 2024-03-08
NR-229763 Add UUID-like header to extensions on all api_server requests

dsellarsnr:


shared/agent/src/api/codestream/codestreamApi.ts
```diff
@@ -384,7 +386,10 @@ export class CodeStreamApiProvider implements ApiProvider {
 		private readonly _version: VersionInfo,
 		private readonly _httpsAgent: HttpsAgent | HttpsProxyAgent<string> | HttpAgent | undefined,
 		private readonly _strictSSL: boolean
-	) {}
+	) {
+		this._clientId = machineIdSync();
```
bcanzanella, line 390: any chance it could throw?

dsellarsnr: Yeah, what should we set it to if it throws - blank?

CalvinAllen (approved):
Pending @bcanzanella's comment

shared/agent/src/api/codestream/codestreamApi.ts
bcanzanella: sounds good

PR #2022, opened 2024-03-11
adds timing descriptor for slow agent requests

bcanzanella:
This will allow us to quickly search the agent logs for "slow" requests. thoughts on the buckets of timings to descriptors?


eric-j (approved):
👍  makes sense to me.  If everything for some reason falls under 500ms, then it might be worth adjusting but I don't think thats the case.

bcanzanella:
> 👍 makes sense to me. If everything for some reason falls under 500ms, then it might be worth adjusting but I don't think thats the case.

yeah, we can adjust the buckets for sure. wasn't sure what the baseline should be, aka is 500ms _ok_ or should that be considered slow? kind of hard to generalize it in the IDE world

PR #2025, opened 2024-03-12
NR-219651: Link to related logs from errors

reneelpetit:


shared/agent/src/providers/newrelic/errors/observabilityErrorsProvider.ts
```diff
@@ -535,6 +536,7 @@ export class ObservabilityErrorsProvider {
 			"latest(message) AS 'message',",
 			"latest(entityGuid) AS 'entityGuid',",
 			"latest(fingerprint) AS 'fingerprintId',",
+			"latest(traceId) AS 'traceId',",
```
bcanzanella, line 539: since there's a timestamp here, that could be passed along to the ApmLog UI to set a default time range that best matches the dropdown options

bcanzanella: (what we talked about in standup)

reneelpetit:
created a different PR https://github.com/TeamCodeStream/codestream/pull/2045

PR #2028, opened 2024-03-12
Visual Studio

CalvinAllen:
1. Polyfill globalThis
2. Polyfill structuredClone
3. Execute them in each webview


dsellarsnr (approved):
Looks good, should work for JB as well

PR #2031, opened 2024-03-12
Make bad queries fail faster

bcanzanella:
used in nrql editor. it will still attempt to do a token refresh (which we could not do, but it's hard to know what that response looks like) so this is the safer path. thoughts?


shared/agent/src/providers/newrelic/newRelicGraphqlClient.ts
```diff
@@ -279,6 +296,11 @@ export class NewRelicGraphqlClient implements Disposable {
 				resp = await client.request<T>(query, variables);
 				// fetchCore will have retried 3 times by now
 			} catch (ex) {
+				if (triedRefresh) {
```
bcanzanella, line 299: could take out the `triedRefresh` to make it fail the first time (wo/checking for token)

dsellarsnr: yes, should be able to take it out - isInvalidInputErrorResponse really narrows it down to having a valid token beforehand in order to get that specific error

bcanzanella: ok, nuked. rdy for your approval.

dsellarsnr (approved):

PR #2052, opened 2024-03-22
Logging

CalvinAllen:
- Adds support for Partitions
- Auto refresh results when criteria changes
- Various issues resolved around timing / re-using windows / etc., both for code search and coming from errors.

### DIFF
--- a/shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
+++ b/shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
@@ -186,7 +186,7 @@ export const APMLogSearchPanel = (props: {
 
 	const [hasPartitions, setHasPartitions] = useState<boolean>(false);
 	const [selectPartitionOptions, setSelectPartitionOptions] = useState<SelectedOption[]>([]);
-	const [selectedPartition, setSelectedPartition] = useState<SelectedOption[]>([]);
+	const [selectedPartitions, setSelectedPartitions] = useState<SelectedOption[]>([]);
 
 	const [originalSearchResults, setOriginalSearchResults] = useState<LogResult[]>([]);
 	const [searchResults, setSearchResults] = useState<LogResult[]>([]);
@@ -496,7 +496,7 @@ export const APMLogSearchPanel = (props: {
 				setHasPartitions(true);
 			}
 
-			setSelectedPartition([defaultPartition]);
+			setSelectedPartitions([defaultPartition]);
 		} catch (ex) {
 			handleError(ex);
 		}
@@ -524,11 +524,11 @@ export const APMLogSearchPanel = (props: {
 				return;
 			}
 
-			const selectedPartitions = defaultPartition
+			const partitions = defaultPartition
 				? [defaultPartition]
-				: selectedPartition.map(p => p.value);
+				: selectedPartitions.map(p => p.value);
 
-			if (selectedPartitions.length === 0) {
+			if (partitions.length === 0) {
 				handleError("Please select at least one partition from the drop down before searching.");
 				return;
 			}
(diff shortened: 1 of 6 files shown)


shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
```diff
@@ -191,57 +204,99 @@ export const APMLogSearchPanel = (props: {
...
+		{ value: "7 DAYS AGO", label: "7 Days Ago" },
+	];
+
+	const defaultPartition: SelectedOption = {
+		value: "Log",
+		label: "Log",
+	};
+
+	useEffect(() => {
+		const debouncedSave = debounce(nextValue => {
```
bcanzanella, line 233: this looks a little weird to me... i _think_ i'd expect the debounced function to be hoisted way up like:


```
const debouncedSave = debounce((nextValue, setQuery) => {
  setQuery(nextValue);
}, 500);

export const APMLogSearchPanel = (props: {
//snip

 useEffect(() => {
    debouncedSave(searchTerm, setQuery);
  }, [searchTerm]);
```
(untested)

CalvinAllen: I guess it could be? - first time I've needed to debounce something in React, and I couldn't find another example in our UI where we've done it on a traditional <input text> and not a drop-down

CalvinAllen: @bcanzanella Why hoist it entirely out of the component, exactly?

shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
```diff
@@ -191,57 +204,99 @@ export const APMLogSearchPanel = (props: {
...
+
+	const sinceOptions: SelectedOption[] = [
+		{ value: "30 MINUTES AGO", label: "30 Minutes Ago" },
+		{ value: "60 MINUTES AGO", label: "60 Minutes Ago" },
+		{ value: "3 HOURS AGO", label: "3 Hours Ago" },
+		{ value: "8 HOURS AGO", label: "8 Hours Ago" },
+		{ value: "1 DAY AGO", label: "1 Day Ago" },
+		{ value: "3 DAYS AGO", label: "3 Days Ago" },
+		{ value: "7 DAYS AGO", label: "7 Days Ago" },
+	];
```
dsellarsnr, line 225: Looks like you don't need the `setSelectSinceOptions` useState from line 184 since this is just a simple lookup that never changes. Can also lift it outside of the component so it isn't (potentially) re-allocated on every re-render.

shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
dsellarsnr: You are creating a new debounceSave every time useEffect runs which means you have no debounce.

shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
CalvinAllen: Ohhhh, good call, yeah that thing never changes (for now)

shared/ui/Stream/APMLogging/APMLogSearchPanel.tsx
CalvinAllen: I see that (now), but it does _work_ and is getting debounced the way I'd expect

CalvinAllen: Hoisted (found another issue in the testing of that)

PR #2061, opened 2024-03-26
NR-250158 Clients in latest 15.5.0 version of extension are looping

dsellarsnr:
Phase 1 of 3 - refactor token storage and refresh logic to single place

tokenHolder centralized storage of access token / refresh token
fetchCore is now single place for performing token refresh

Future phases:
Phase 2 - Detect hard loops, gather and send logs, logout session

Phase 3 - Stop updating preferences when there is no update


bcanzanella (approved):
🚀

PR #2063, opened 2024-03-27
Pull in list of display names for service types and use in log search

eric-j:
https://new-relic.atlassian.net/browse/NR-251885


shared/agent/src/providers/newrelic/entityTypeDisplayNames.ts
```diff
@@ -0,0 +1,5087 @@
...
+		domain: "SYNTH",
+		id: "SYNTH-SECURE_CRED",
+		type: "SECURE_CRED",
+		uiDefinitions: {
+			displayName: "Secure credential",
+			displayNamePlural: "Secure credentials",
+		},
+	},
+	{
+		domain: "TEST",
```
bcanzanella, line 4941: can nuke these test ones

bcanzanella: actually, if we're going to add another layer for customizations, maybe then leave these here (since subsequent updates would need these removed too, probably easiest to just leave them)

eric-j: Yeah I'm just going to leave them given the array will eventually be updated.

PR #2067, opened 2024-03-29
NR-235249: Git error trying to use instrumentation wizard

reneelpetit:
https://new-relic.atlassian.net/browse/NR-235249


eric-j (approved):
👍

PR #2092, opened 2024-04-05
NR-251900: "codestream/vulnerability_link clicked" tracking called on every render

reneelpetit:
https://new-relic.atlassian.net/browse/NR-251900


dsellarsnr (changes requested):
The original problem still remains - 
image

the rendering part of a react component will re-run anytime something changes so this code block still gets executed hundreds of times. You probably need to put this inside of a `useDidMount` and then test it to make sure the event only fires once.

dsellarsnr (approved):

PR #2095, opened 2024-04-09
APM Logs

CalvinAllen:
* Add "informational" message in addition to "error", and set entity message as such
* Add new meta_data property to telemetry that adds the new entity display name


eric-j (approved):
👍

PR #2154, opened 2024-05-20
VS: Fix Anomaly Objects

CalvinAllen:
Mostly just object updates to match agent


CalvinAllen:
And the event to handle forwarding the agent notification to the webview

dsellarsnr (approved):

PR #2162, opened 2024-05-29
Don't change user readable values in front-end, so tack on "AGO" into the query and remove it from the fallback/default.

CalvinAllen:
Keep in line with Prod, so users see "3 days", "7 days", but missing "AGO" when placed into template so just hard-code it there.

### DIFF
--- a/shared/agent/src/providers/newrelic/errors/errorQueries.ts
+++ b/shared/agent/src/providers/newrelic/errors/errorQueries.ts
@@ -194,7 +194,7 @@ const errorQueryGroups: ErrorQueryTemplate[] = [
 				"FROM ErrorTrace",
 				`WHERE fingerprint IS NOT NULL AND NOT error.expected AND entityGuid='<%= applicationGuid %>'`,
 				"FACET error.class, message", // group the results by identifiers
-				`SINCE <%= since %>`,
+				`SINCE <%= since %> AGO`,
 				"LIMIT MAX",
 			].join(" ")
 		),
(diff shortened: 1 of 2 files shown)


shared/agent/src/providers/newrelic/errors/errorQueries.ts
```diff
@@ -194,7 +194,7 @@ const errorQueryGroups: ErrorQueryTemplate[] = [
 				"FROM ErrorTrace",
 				`WHERE fingerprint IS NOT NULL AND NOT error.expected AND entityGuid='<%= applicationGuid %>'`,
 				"FACET error.class, message", // group the results by identifiers
-				`SINCE <%= since %>`,
+				`SINCE <%= since %> AGO`,
```
bcanzanella, line 197: this seems weird, like something changed upstream, is there risk of <%=since => also having the "AGO" ? any need to guard against AGO AGO?

CalvinAllen: From my research, no, as that comes from a hard-coded list in another area (no changes there so its not in the PR). Upon examining the preference that gets saved from that setting, it also didn't have the "ago" as part of it.

CalvinAllen: In `api.protocol.models.ts` -

```typescript
export enum CodeErrorTimeWindow {
	"HalfHour" = "30 minutes",
	"Hour" = "1 hour",
	"ThreeHours" = "3 hours",
	"SixHours" = "6 hours",
	"TwelveHours" = "12 hours",
	"OneDay" = "24 hours",
	"ThreeDays" = "3 days",
	"SevenDays" = "7 days",
}
```

bcanzanella: got it, thanks for checking.

bcanzanella (approved):

PR #2170, opened 2024-06-04
Stop the comment replacement regex from messing with things that have forward slashes when they have been single quoted

CalvinAllen:
This took longer than I had hoped...

While trying to figure out a way to escape the slashes, it finally clicked that the regex for handling comments was being too aggressive. Now we only remove "unquoted" comments, allowing things like URLs, etc., to be skipped as a string constant, not a comment.

One new test added, and _all_ tests are passing (so it hopefully didn't break existing comment replacement functionality)

### DIFF
--- a/shared/agent/test/unit/providers/newrelic/nrqlProvider.spec.ts
+++ b/shared/agent/test/unit/providers/newrelic/nrqlProvider.spec.ts
@@ -101,12 +101,10 @@ AND status = 'baz'
 		expect(result).toBe("");
 	});
 
-	it("escape forward slashes", () => {
+	it("treat single-quoted slashes as string literals and not remove them", () => {
 		const result = provider.transformQuery(`FROM Collection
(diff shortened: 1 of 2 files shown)


shared/agent/test/unit/providers/newrelic/nrqlProvider.spec.ts
```diff
@@ -100,6 +100,16 @@ AND status = 'baz'
 */`);
 		expect(result).toBe("");
 	});
+
+	it("escape forward slashes", () => {
```
dsellarsnr, line 104: rename for clarity - should it be "doesn't treat quoted URLs as comments"?

CalvinAllen: Ah, yeah, something like that - think I forgot to change it after I switched up how I was implementing the fix

PR #2179, opened 2024-06-11
fix for closing dialog

eric-j:


shared/ui/Stream/ErrorRoadblock.tsx
```diff
@@ -53,7 +54,7 @@ export const ErrorRoadblock = props => {
 	};
 
 	return (
-		<Dialog title="Cannot open Error" onClose={() => dispatch(closeModal())}>
+		<Dialog title="Cannot Open Error" onClose={() => dispatch(closeModal())}>
```
CalvinAllen, line 57: Should this be bound to `onSubmit` as well?

CalvinAllen (approved):
1 question / comment

shared/ui/Stream/ErrorRoadblock.tsx
eric-j: No, `<Dialog />` has its own onClose prop.  It might be a bit redundant but I'd like to keep the same as other Dialogs

PR #2187, opened 2024-06-18
NR-280242 Looping requests to nerdgraph

dsellarsnr:


shared/agent/src/rateLimits.ts
```diff
@@ -23,11 +15,12 @@ const codestreamOrigins = new Set([
 	"https://codestream-us1.service.newrelic.com",
 ]);
```
CalvinAllen, line 16: No EU?

shared/agent/src/rateLimits.ts
```diff
@@ -23,11 +15,12 @@ const codestreamOrigins = new Set([
...
 
-// Rate limits
+// Rate limits - this is per RATE_LIMIT_INTERVAL seconds
+// warn: log a warning, but don't block
+// report: log an error (errors inbox), but don't block
+// block: block the request
```
CalvinAllen, line 21: I don't see block being used - outdated comment, or missing in code?

shared/agent/src/rateLimits.ts
dsellarsnr: line 14

shared/agent/src/rateLimits.ts
dsellarsnr: It is still a possibility - just not used at the moment. Wanted to make it more visible since it not used in a rule at the moment.

dsellarsnr: If you block before you forceLogout, forceLogout never happens

shared/agent/src/rateLimits.ts
CalvinAllen: Ah, the viewer cut a bunch of lines, I didn't catch it.

PR #2191, opened 2024-06-20
Inline Create Icons can't appear if I delete them all.

CalvinAllen:
Half of these files are dead anyway, but had to fix them to resolve compilation errors


eric-j (commented):
👍

PR #2194, opened 2024-06-27
ux and telemetry bug fixes for notifications menu

eric-j:


shared/ui/Stream/Notifications.tsx
```diff
@@ -24,6 +24,7 @@ const NotficationSubHeaders = styled.div`
 
 export const Notifications = props => {
 	const dispatch = useAppDispatch();
+	const elementRef = useRef(null);
```
dsellarsnr, line 27: `const elementRef = useRef<HTMLFormElement>(null);`
to get rid of the `//@ts-ignore`

PR #2196, opened 2024-06-28
Feature/pubnub to v3

cstryker:
Enables extension to continue to use V2 PubNub tokens (which are issued by our server then associated with access permissions via PubNub), but also be ready to use V3 PubNub tokens (which are issued by PubNub with the requested permissions embedded). This should allow a seamless switch from the legacy API server to the new one.


dsellarsnr (approved):

dsellarsnr:
Obsolete PR

PR #2197, opened 2024-06-28
NR-284560 Investigate spikes of calls to GET /posts in staging

dsellarsnr:
- Fix unable to open codeerrors: Codemarks can never be destroyed


eric-j (approved):
👍 Looks like mostly just adding back deleted code?

marcelofarias (approved):

PR #2199, opened 2024-07-01
Ensure we don't end up with blank screen if code error is missing stack trace

purplexa:
The changes look like a bit but it's just moving the if statement from line 325 in the original to line 424 in the new, but that changes the indentation and the diff processing can't figure it out apparently.


eric-j (approved):
👍

dsellarsnr (approved):

PR #2214, opened 2024-07-17
NR-292683

CalvinAllen:
- Stub out logic for closing an errors discussion thread


reneelpetit (approved):
I find "kebab-horizontal" to be a funny name :)  Thinking of kebabs makes me hungry.

Looks good!

PR #2215, opened 2024-07-17
Update logic for determining NRAI access

purplexa:


shared/agent/src/providers/newrelic/newRelicGraphqlClient.ts
```diff
@@ -363,7 +373,10 @@ export class NewRelicGraphqlClient implements Disposable {
 	async query<T = any>(
 		query: string,
 		variables: any = undefined,
-		isMultiRegion = false
+		isMultiRegion = false,
+		extraValues?: {
+			unsafeExperimentalNamespaces?: string[];
```
CalvinAllen, line 378: I don't know how much it matters, but with the work I did for code errors, you can now pass in a new header into the graphql client in the constructor of the specific provider (see `discussions.provider.ts` and it persists for every query that provider makes.

purplexa: oh, that's probably better than this mess. I'll fix it

CalvinAllen (dismissed):
forgot to approve

dsellarsnr:
@purplexa do we need this for tomorrow's release?

purplexa:
@dsellarsnr I don't think so. The existing logic should work for preview customers, and last I heard it wasn't GA yet. This just adds the logic necessary for GA

CalvinAllen (approved):

CalvinAllen (approved):

PR #2227, opened 2024-07-23
NR-293802

CalvinAllen:
1. Ditch weird left-border on comments
2. Fix margin of code-fix to match comments


dsellarsnr (changes requested):
bar-left-parent css is unused now - might as well remove it (~line 137)

dsellarsnr (approved):

PR #2232, opened 2024-07-23
Make codeBlock optional for NRAI calls

CalvinAllen:


shared/ui/Stream/CodeError/CodeError.tsx
```diff
@@ -270,12 +272,12 @@ export const CodeError = (props: CodeErrorProps) => {
 		}
 
 		// grok no good
-		if (!showGrok || derivedState.functionToEditFailed || !derivedState.functionToEdit) {
```
dsellarsnr, line 273: These checks are still needed but they probably need some fixin. If you don't check functionToEditFailed you will end up sending the request before the request to get the code block has finished - it takes awhile. 

functionToEditFailed: true
finished attempt to get code block but it failed. proceed with initializeNrAiAnalysis. 

derivedState.functionToEdit: true
finished attempt to get code block with success, proceed with initializeNrAiAnalysis.

dsellarsnr (approved):

PR #2241, opened 2024-07-25
If we have NO lines and therefore never enter the method to find anything to jump to, we need to dispatch the failure scenario

CalvinAllen:


shared/ui/Stream/CodeError/CodeError.tsx
```diff
@@ -112,6 +112,8 @@ export const CodeError = (props: CodeErrorProps) => {
 		if (stackTraceHasUserLines && !selectedLineIndex) {
 			initializeStackTrace();
 			scrollToStackTrace();
+		} else {
+			dispatch(setFunctionToEditFailed(true));
```
dsellarsnr, line 116: Not sure this will work - testing it out now. I think stackTraceLines might start with no lines resolved until the IDE has a chance to send back the files it finds so this might set failed before the stack trace has a chance to resolve user lines.

dsellarsnr: Won't work - does immediately set failed before checking for user lines. I think I dealt with this last time with a timer waiting for stackTraceLines to be idle for a couple seconds before checking for no user lines. 

image

PR #2243, opened 2024-07-25
NR-295732

CalvinAllen:
Make sure we only process mentions and files through the mentions method - had to add back in some brittleness to the regex, but otherwise we get caught in an infinite loop and the agent crashes.

### DIFF
--- a/shared/agent/src/providers/newrelic/discussions/discussions.provider.ts
+++ b/shared/agent/src/providers/newrelic/discussions/discussions.provider.ts
@@ -48,7 +48,7 @@ import * as htmlparser2 from "htmlparser2";
 @lsp
 export class DiscussionsProvider {
 	// TODO fix brittleness - relies on the properties of the collab-mention tag being in the right order...
-	private collabTagRegex = /<collab-mention data-type="(NR_USER|FILE)".*?<\/collab-mention>/ims;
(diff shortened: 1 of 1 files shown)


shared/agent/src/providers/newrelic/discussions/discussions.provider.ts
```diff
@@ -48,7 +48,7 @@ import * as htmlparser2 from "htmlparser2";
 @lsp
 export class DiscussionsProvider {
 	// TODO fix brittleness - relies on the properties of the collab-mention tag being in the right order...
-	private collabTagRegex = /<collab-mention.*?<\/collab-mention>/ims;
+	private collabTagRegex = /<collab-mention data-type="(NR_USER|FILE)".*?<\/collab-mention>/ims;
```
dsellarsnr, line 51: `/<collab-mention .*data-type="(NR_USER|FILE)".*?<\/collab-mention>/ims` - add a `.*` so you can find data-type in any position

CalvinAllen: Done

dsellarsnr (dismissed):

CalvinAllen:
Waiting for CI just to be certain that tests still pass...

PR #2247, opened 2024-07-25
Update server URLs

purplexa:
I verified that the url migration worked in VSC and JB, but not VS since I don't know anything about it


CalvinAllen (approved):
VS looks good

dsellarsnr (approved):

PR #2258, opened 2024-07-30
"Assigned To Me" errors section only goes back 30 minutes

reneelpetit:
I added in the time option, but it doesn't seem that the different time frame queries are going through for either the recent errors section or the assigned to me section. it just shows the same ones regardless of whether you select 24 hours or 7 days. Working on fixing that


eric-j (dismissed):
👍

reneelpetit:
it seems to be working as is without any changes needed

PR #2269, opened 2024-08-06
Ensure Recent Errors links work correctly for all time windows

purplexa:


eric-j (dismissed):
👍

shared/agent/src/providers/newrelic/errors/observabilityErrorsProvider.ts
```diff
@@ -1354,4 +1371,22 @@ export class ObservabilityErrorsProvider {
 			}
 		);
 	}
+
+	private adjustErrorGroupUrl(url: string, timeWindow: string): string {
+		const timeWindowToMs: Record<string, number> = {
```
dsellarsnr, line 1376: move timeWindowToMs outside the class - creating a new Map here every time eats in to the performance gains of using a lookup table.

dsellarsnr (approved):

PR #2276, opened 2024-08-14
fix for ai mention and remove assignee

eric-j:


shared/ui/Stream/RequestTypeHooks/useUserSearch.tsx
```diff
@@ -53,16 +57,28 @@ export const useUserSearch = () => {
 					}
 					setUserSearchResults([]);
 				}
+			} else if (_query === "ai" && mappingStyle === "mentions") {
```
CalvinAllen, line 60: Does this mapping style already get passed in somewhere else?

CalvinAllen (approved):

shared/ui/Stream/RequestTypeHooks/useUserSearch.tsx
eric-j: this line in the mentionstextinput component:

```	const debouncedFetchUsers = useCallback(
		_debounce((query, callback) => fetchUsers(query, "mentions", callback), 300),
		[]
	);```

